PR #30, opened 2026-04-11
Smarter needs_review state: detect existing PR, offer review dispatch

### LINKED ISSUE #27: Smarter needs_review state: detect existing PR, offer review dispatch
~~When a work item moves to needs_review, the panel always shows 'Push & create PR' regardless of context.~~

**Mostly done.** The post-actions system (#31, merged) handles this dynamically now. Skills declare their post-actions via frontmatter, and the Panel renders the right buttons.

**Remaining from PR #30 (still open):**
- PR detection: check if a PR exists for the branch and show 'View PR' instead of 'Push & create PR'
- Store pr_url on the work item model for instant lookups
- Migration system for schema evolution

PR #30 can be reviewed and merged to complete this.

jasonmadigan:
When a work item moves to needs_review, the panel always shows 'Push & create PR'. But if a PR already exists (created by the agent, by push-and-pr, or manually), this is wrong.

The panel should detect whether a PR exists for the work item's branch and show the right actions:

**No PR exists:**
- Push & create PR button (current behaviour)

**PR exists:**
- View PR link
- 'Dispatch review' button (pre-selects review-pr skill, feeds PR URL/diff as context)
- 'Dispatch changes' button (pre-selects address-feedback skill, for when the PR has review comments)

Detection: check if a PR exists on the `clawdio/<item-id>` branch via `gh pr list --head <branch>` or by tracking the PR URL in the work item after creation.

Could also store the PR URL on the work item model so we don't need to check every time. The push-and-pr endpoint already returns the URL -- just need to persist it.

Closes jasonmadigan/clawdio#27


jasonmadigan (commented):
a sentence with a full stop (e.g. `already exists: https://github.com/foo/bar/pull/1.`), it'll capture the trailing `.` or `'`. In practice `gh` currently outputs clean URLs, but it's fragile. Trimming trailing punctuation from the match would be safer.

### Minor

**Comment says "verify it's still valid" but doesn't** -- `handlers.go:496-498`

The comment on `handleCheckPR` says "already have a stored PR URL -- verify it's still valid" but the code just returns it without verification. The comment is misleading; either verify or fix the comment.

**Migration errors silently swallowed** -- `store.go:41`

`db.Exec(m)` with no error check is intentional (column already exists = expected failure), but if the migration fails for a *different* reason (disk full, permissions), you won't know. The pattern is fine for a single `ALTER TABLE ADD COLUMN`, but as the `migrations` slice grows, you'll want to distinguish "column exists" from actual failures. A `TODO` or checking the specific error string would help future-proof this.

**`handleCheckPR` uses worktree path that may not exist** -- `handlers.go:519-525`

If the worktree directory has been cleaned up (manually or by the system), `gh pr list` with `cmd.Dir = wtPath` will fail with a working directory error. The code handles this (returns empty pr_url), so it won't crash, but the user gets no feedback about *why* it failed. Not a real issue today since the `gh pr list --head` command doesn't actually need to run from the repo directory if the remote is configured globally, but worth noting.

## Suggestions

1. Change `"status": "created"` to `"status": "exists"` in the "PR already exists" path for semantic accuracy.
2. Log `UpdateWorkItemPRURL` errors at warn level -- a one-liner each.
3. Fix the "verify it's still valid" comment to match what the code does.

## Verdict

**Approve.** Clean, well-scoped change that solves the problem described in #27. The issues above are all minor/nice-to-have -- nothing blocks merging.

PR #36, opened 2026-04-11
Add CI: build + test + typecheck on every PR

### LINKED ISSUE #16: Add CI: build + test + typecheck on every PR
We merged a PR that broke the skill manager call site -- the function signature changed but the caller wasn't updated. A basic CI pipeline would have caught this.

Add a GitHub Actions workflow that runs on every PR:

```yaml
- go build ./...
- go test ./...
- cd ui && npx tsc --noEmit
```

This is the minimum. Could also add linting (golangci-lint, eslint) later.

jasonmadigan:
We merged a PR that broke the skill manager call site -- the function signature changed but the caller wasn't updated. A basic CI pipeline would have caught this.

Add a GitHub Actions workflow that runs on every PR:

```yaml
- go build ./...
- go test ./...
- cd ui && npx tsc --noEmit
```

This is the minimum. Could also add linting (golangci-lint, eslint) later.

Closes jasonmadigan/clawdio#16


jasonmadigan (commented):
Here's my review:

---

## Summary

Adds a GitHub Actions CI workflow that runs Go build, Go test, and TypeScript typechecking on PRs targeting main. Single file addition, no code changes. Directly addresses the issue where a broken call site was merged without being caught.

## Findings

No critical or important issues. The workflow is clean and straightforward.

**Minor:**

- `.github/workflows/ci.yml` -- No `push` trigger on `main`. This means direct pushes to main (if branch protection doesn't prevent them) won't run CI. Probably fine if branch protection is in place, but worth noting.

- The workflow doesn't run on the `push` event for the PR branch itself, which means force-pushes after the initial PR open will still trigger (that's fine -- `pull_request` covers synchronize events by default).

## What's good

- `go-version-file: go.mod` rather than hardcoding a version -- picks up the project's Go version automatically.
- `npm ci` instead of `npm install` -- correct for CI (deterministic, faster).
- `cache: npm` with the right `cache-dependency-path` -- avoids redundant installs.
- Single job keeps it simple. No over-engineering.
- Pinned to `@v4`/`@v5` action versions.

## Verdict

**Approve.** This does exactly what it says on the tin. The workflow is minimal, correct, and would have caught the breakage that motivated it. Linting can come later as mentioned in the PR description.

jasonmadigan (commented):
Here's my review:

---

## CI Status

**Failing.** `build-and-test` fails at the Go test step. Two tests fail in `internal/skill`:

- `TestSourceForDir` -- `expected claude-code, got clawdio`
- `TestListIncludesAllSources` -- `missing source "claude-code" in skills`

## Summary

Adds a GitHub Actions workflow that runs `go build`, `go test`, and TypeScript typechecking on PRs targeting main. Single file, 33 lines, clean and minimal.

## Findings

**Important: Pre-existing test bug surfaced by CI**

The failing tests aren't caused by this PR -- they exist on `main` too. The bug is in `sourceForDir` (`internal/skill/skill.go:235`): the condition `strings.HasSuffix(dir, "/skills")` matches *both* `~/.clawdio/skills` and `~/.claude/skills`, so all skill directories get labelled `"clawdio"`. The `.claude/skills` case never reaches the `return "claude-code"` fallback.

This is the exact kind of thing CI should catch, so good news on that front. But the branch needs a rebase onto main (it's 7 commits behind) and that test bug needs fixing before this can merge cleanly.

**Minor: No `push` trigger for main**

The workflow only runs on `pull_request`. Consider also running on pushes to `main` so you catch breakage from direct commits or merge conflicts. Common pattern:

```yaml
on:
  push:
    branches: [main]
  pull_request:
    branches: [main]
```

Not blocking, just a suggestion.

## Suggestions

1. Rebase onto current `main`.
2. Fix the `sourceForDir` bug -- the `.clawdio` check needs to come first and the `/skills` suffix check should be removed or scoped properly (e.g. check the parent directory, not just the suffix).
3. Consider adding `push` trigger for `main`.

## Verdict

**Request changes** -- CI is failing. The workflow itself is sound, but pre-existing test failures need resolving (either fix the bug or fix the tests) and the branch needs rebasing.

jasonmadigan (commented):
branch exists** (`internal/worktree/worktree_test.go`, lines 169/199/252/293 on main). The `setupTestRepoWithRemote` helper does `git push -u origin HEAD` which pushes whatever the default branch name is (often `master` on Ubuntu CI runners). The tests then call `git checkout main` which fails with `pathspec 'main' did not match any file(s) known to git`. This is a pre-existing bug on main, not introduced by this PR, but the CI workflow is now surfacing it. Fix options:
   - Add `git init -b main` or `git config init.defaultBranch main` in the test helper
   - Or rename the bare repo's HEAD to main after init

**Important**

2. **`sourceForDir` fallback changed** (`internal/skill/skill.go:241`). The else-fallback changed from `"claude-code"` to `"clawdio"`. Before: unknown dirs defaulted to `claude-code`. After: unknown dirs default to `clawdio`. This seems intentional (`.claude` now has its own check first), but the effect is that any skill directory that isn't `.claude` or `.clawdio` will now be labelled `clawdio` instead of `claude-code`. Worth confirming that's the desired behaviour -- are there any other skill directories that should still map to `claude-code`?

**Minor**

3. **No branch protection check.** The workflow runs on `push` to main and PRs targeting main, which is right. No issues here, just noting that the `push: branches: [main]` trigger means merges to main also run CI, which is fine for validation.

### Suggestions

- Fix the worktree tests by setting `init.defaultBranch` to `main` in `setupTestRepoWithRemote`. Something like adding `{"git", "config", "init.defaultBranch", "main"}` to the bare repo init, or using `git init --bare -b main`.
- Consider whether the `skill.go` change belongs in this PR or should be a separate commit. It's unrelated to CI.

### Verdict

**Request changes.** CI is failing. The worktree test issue is pre-existing but needs fixing for CI to pass. The `skill.go` change is orthogonal and should be confirmed as intentional.

jasonmadigan (commented):
ere `~/.claude/skills` was incorrectly labelled as `clawdio`.

### Findings

**Critical**

1. **`internal/worktree/worktree_test.go` was truncated.** The file on this branch is 97 lines; on main it's 351. The PR intended to change only line 15 (`git init` -> `git init -b main`) but instead deleted everything after `TestCreateIdempotent`: `setupTestRepoWithRemote`, `TestCreateFromBranch`, `TestCreateFromBranchUpdatesExisting`, `TestCreateFromBranchConflict`, `TestUpdateBranch`, and `TestBranchInWorktree` are all gone. The remaining `setupTestRepoWithRemote` helper also uses `git init --bare` which needs the same `-b main` fix. This is what's causing the CI failure -- the deleted tests still exist on main and run against the unfixed `setupTestRepoWithRemote`.

   Fix: restore the full test file from main, then apply `-b main` to both `setupTestRepo` (line 15) and `setupTestRepoWithRemote`'s bare init (line 56) and its clone setup.

**Minor**

2. **`sourceForDir` fallback changed from `"claude-code"` to `"clawdio"` (skill.go:241).** The old fallback was `"claude-code"` for unknown dirs. Now both the `.clawdio`/`skills` branch and the fallback return `"clawdio"`. This means any directory that doesn't match `.claude` or `.clawdio` patterns now defaults to `"clawdio"` instead of `"claude-code"`. Probably fine since the function is only called on known skill directories, but worth a conscious check that no other callers rely on the old default.

3. **The `sourceForDir` fix is unrelated to the CI workflow.** It's a legitimate bugfix but should ideally be in its own commit or PR for clarity. It's already in a separate commit (`3b6cc6f`) which is acceptable.

### Suggestions

- Restore the deleted test functions in `worktree_test.go` and fix both `git init` calls.
- Consider adding a `"skills"` case to `TestSourceForDir` for the bare `"skills"` path.

### Verdict

**Request changes.** CI is failing because of truncated test file. The workflow itself looks correct and minimal.

PR #39, opened 2026-04-11
Sync work item descriptions when GitHub issues are updated

### LINKED ISSUE #38: Sync work item descriptions when GitHub issues are updated
The GitHub adapter uses a `seen` map that marks items after first import and skips them forever. This means:

- Issue descriptions edited on GitHub are never updated in clawdio
- New comments on issues aren't reflected
- Label changes aren't picked up
- Only open/closed status is synced (via `CheckStatus`)

**Root cause:** `Poll()` checks `g.seen[ref]` and skips any item already imported. The method only returns genuinely new items.

**Fix options:**

1. **Re-fetch and diff:** On each poll, re-fetch all items (not just unseen). Compare `updated_at` or body hash against what's in the store. Update if changed. This is more API calls but `gh` handles rate limiting.

2. **Separate sync pass:** After polling for new items, do a second pass over existing items checking for updates. Use `gh issue view <number> --json updatedAt,body` and compare.

3. **Webhook-based:** Instead of polling, use GitHub webhooks to push updates. More complex but real-time. Probably overkill for v1.

**Recommendation:** Option 1 is simplest. Change `Poll()` to return both new items and updates. The caller in main.go then does `CreateWorkItem` for new and `UpdateWorkItem` for existing. The `seen` map becomes a `known` map that tracks the last-seen body hash, not just existence.

jasonmadigan:
The GitHub adapter uses a `seen` map that marks items after first import and skips them forever. This means:

- Issue descriptions edited on GitHub are never updated in clawdio
- New comments on issues aren't reflected
- Label changes aren't picked up
- Only open/closed status is synced (via `CheckStatus`)

**Root cause:** `Poll()` checks `g.seen[ref]` and skips any item already imported. The method only returns genuinely new items.

**Fix options:**

1. **Re-fetch and diff:** On each poll, re-fetch all items (not just unseen). Compare `updated_at` or body hash against what's in the store. Update if changed. This is more API calls but `gh` handles rate limiting.

2. **Separate sync pass:** After polling for new items, do a second pass over existing items checking for updates. Use `gh issue view <number> --json updatedAt,body` and compare.

3. **Webhook-based:** Instead of polling, use GitHub webhooks to push updates. More complex but real-time. Probably overkill for v1.

**Recommendation:** Option 1 is simplest. Change `Poll()` to return both new items and updates. The caller in main.go then does `CreateWorkItem` for new and `UpdateWorkItem` for existing. The `seen` map becomes a `known` map that tracks the last-seen body hash, not just existence.

Closes jasonmadigan/clawdio#38


jasonmadigan (commented):
No unique constraint on `source_ref`, and `GetWorkItemBySourceRef` uses `QueryRow` which returns the first match. Let me also check the `contentHash` separator choice and the notification handling.

I've now reviewed everything I need. Here's the review:

---

## PR Review: Sync work item descriptions when GitHub issues are updated (#39)

### CI Status

Passing.

### Summary

Replaces the boolean `seen` map in the GitHub adapter with a `known` map that tracks content hashes (`sha256(title + "\x00" + body)`). `Poll()` now returns a `PollResult` struct separating new items from updates. Adds `SyncUpdates` to the poller and `GetWorkItemBySourceRef` to the store.

### Findings

**Important:**

1. **`SyncUpdates` is never called.** The PR adds `Poller.SyncUpdates()` and changes `Poll()` to return `PollResult` with separate `New`/`Updated` slices, but no caller in the PR invokes `SyncUpdates(result.Updated)`. The entrypoint (`cmd/clawdio/main.go`, referenced in CLAUDE.md as "the most important file") isn't in the repo at all -- it seems to exist outside version control. Without the caller wiring, this PR is inert: updates are detected by the adapter but silently discarded. Whoever wires this up needs to be aware.

2. **`source_ref` has no unique constraint.** `GetWorkItemBySourceRef` uses `QueryRow` which returns the first row matching the WHERE clause. If duplicate `source_ref` values exist (e.g. a manual item and a GitHub import with the same ref, or a bug creates duplicates), only one gets updated. Consider adding a `UNIQUE` index on `source_ref` in a migration, or at minimum `LIMIT 1` with an `ORDER BY` to make behaviour deterministic. Not a blocker for this PR but worth a follow-up.

3. **Notifications never produce updates.** `pollNotifications` uses the old pattern: `if _, exists := g.known[ref]; exists { continue }` with an empty hash `""`. It can never detect content changes to notifications, only new ones. The PR description says "new comments on issues aren't reflected" as a goal, but notification content changes won't be caught. This is probably fine since notifications are ephemeral signals, but worth documenting the intentional asymmetry.

**Minor:**

4. **`contentHash` separator choice.** Using `title + "\x00" + body` is fine, but `"\x00"` can appear in issue bodies (e.g. pasted binary data, though rare). If this ever caused a collision it would just mean a missed update, not data corruption. Acceptable risk.

5. **Label changes still not synced.** The PR description mentions "label changes aren't picked up" as a problem, but `contentHash` only hashes title and body. Labels aren't factored in. If labels matter for work item type or status, they'll still be stale. Not a regression though -- just an unaddressed part of the issue.

6. **Comment in `pollIssues` removed silently.** The `// best effort for notifications` comment was removed from `Poll()` at line 143 of the old code. The behaviour is the same (still `continue` on error), but the rationale is gone.

### Suggestions

- Wire up the `SyncUpdates` call in the entrypoint before merging, or split this into two PRs: one for the plumbing (this PR) and one for the integration. Currently the feature doesn't work end-to-end.
- Consider adding a `UNIQUE` constraint on `source_ref WHERE source_ref != ''` as a follow-up.

### Verdict

**Request changes.** The core mechanism is sound and well-tested, but the feature is incomplete: `SyncUpdates` is dead code without the caller integration. Either add the caller or clarify in the PR description that this is plumbing-only and the integration comes in a follow-up.

jasonmadigan (commented):
Good -- `QueryRow` only returns one row, so if duplicates exist it silently picks the first. That's an existing concern, not introduced by this PR.

Now I have all the context I need. Here's my review:

---

## PR Review: Sync work item descriptions when GitHub issues are updated (#39)

### CI Status

Passing.

### Summary

Replaces the `seen` map (tracks existence) with a `known` map (tracks content hashes). `Poll()` now returns a `PollResult` with separate `New` and `Updated` slices. Adds `SyncUpdates` to the poller and `GetWorkItemBySourceRef` to the store. Notifications are correctly excluded from update tracking since they're ephemeral signals.

### Findings

**Important:**

1. **`SyncUpdates` and `MarkKnown` are not wired into `main.go`.** The commit message says "wire SyncUpdates into pollOnce (in untracked main.go)" but `cmd/clawdio/main.go` is not in this diff. This means the feature is built but not activated. If `main.go` lives outside this worktree and is meant to be a separate change, this should be documented. Otherwise the `Updated` slice is produced but never consumed, and `MarkKnown` is never called at startup so every item will appear as new on first poll after restart.

2. **`source_ref` has no unique index.** `GetWorkItemBySourceRef` uses `QueryRow` which silently returns the first match if duplicates exist. Nothing currently prevents two work items with the same `source_ref` (e.g. if a user manually creates one and the adapter imports the same issue). Only one would get updated. This is a pre-existing concern but worth noting now that the column is load-bearing for update logic -- consider adding a unique index in a follow-up.

3. **Hash separator collision.** `contentHash` uses `title + "\x00" + body`. A title ending in `\x00` and a different body could produce the same hash as a different title/body pair. Extremely unlikely in practice with GitHub content, but using a length-prefixed scheme or double separator would be more robust. Not a blocker.

**Minor:**

4. **Notifications store empty hash:** `g.known[ref] = ""` for notifications. This is fine for the current "skip if seen" logic, but it's a different semantic than the issue/PR entries. A comment or named constant would make the intent clearer. The existing comment on `pollNotifications` partially covers this.

5. **`SyncUpdates` only syncs title and description.** Labels, assignee, and other metadata mentioned in the issue description aren't synced. This is fine for v1 but should be noted as a known limitation.

### Suggestions

- Wire up `SyncUpdates` and `MarkKnown` in `main.go` as part of this PR, or open a follow-up issue tracking that integration. Shipping the plumbing without the wiring means the feature doesn't work yet.
- Consider `LIMIT 1` on the `GetWorkItemBySourceRef` query to make the single-row intent explicit.

### Verdict

**Request changes** -- the core logic is solid and well-tested, but the feature isn't wired in. The `main.go` changes should either be included here or a follow-up issue should track them. Without that, the `Updated` slice is produced and discarded.

jasonmadigan (commented):
The `contentHash` for PRs uses `pr.Title` (without the `[PR]` prefix), but the work item's `Title` includes it. So the hash is computed from the raw GitHub title, while the stored title has the prefix. When `SyncUpdates` compares `existing.Title == u.Title`, the existing title is `[PR] foo` and the update carries `[PR] foo` — so that's fine, the prefix is added before the item goes into the result. The hash comparison is consistent because it always uses the raw title from GitHub. Good.

Let me check one more thing — whether `UpdateWorkItem` could accidentally overwrite fields like `skill_name` or `profile_id` that shouldn't change during a sync.

The `SyncUpdates` method at `poller.go:95-107` fetches the full existing item, only overwrites `Title` and `Description`, then calls `UpdateWorkItem` which writes all fields. Since it fetches the existing item first, the other fields are preserved. Clean.

Here's my review:

---

## CI Status

Passing.

## Summary

Replaces the boolean `seen` map with a content-hashing `known` map so that GitHub issue/PR description and title changes are detected on subsequent polls and synced to the local store. Adds `GetWorkItemBySourceRef` to the store, `SyncUpdates` to the poller, and `PollResult` to separate new items from updates. Also fixes `.gitignore` to anchor `/clawdio` so it doesn't ignore the `cmd/clawdio/` source directory.

## Findings

### Important

1. **`source_ref` has no unique constraint** — `GetWorkItemBySourceRef` uses `LIMIT 1` which is fine defensively, but if duplicate `source_ref` values exist (e.g. from a bug or manual creation), the "wrong" item could be updated silently. This is a pre-existing schema issue, not introduced by this PR, but worth noting since this query is now load-bearing for sync correctness. A unique index on `(source, source_ref)` would be a good follow-up.

2. **`known` map is not goroutine-safe** — `pollOnce()` is launched via `go pollOnce()` on startup (line 245) and also called from a ticker goroutine (line 249). If the first poll takes longer than the ticker interval, both goroutines write to `gh.known` concurrently. Pre-existing issue (the old `seen` map had the same problem), and unlikely with a 2-minute default interval, but a `sync.Map` or mutex would make it correct. Again, fine as a follow-up.

### Minor

3. **Notification hash is empty string** — `github.go:281` sets `g.known[ref] = ""` for notifications, which is intentional and documented ("notifications are ephemeral signals so content changes don't apply"). Just confirming this is deliberate — the comment at line 256 explains it well.

4. **`MarkKnown` no-op on existing entries** — the guard at `github.go:96` (`if _, ok := g.known[ref]; !ok`) means if an item's content changed in the DB between restarts (e.g. manual edit), `MarkKnown` won't pick up the DB version. This is fine for the current use case (seeding at startup prevents spurious updates) but worth knowing.

## Suggestions

- Consider adding a unique index on `(source, source_ref)` as a follow-up to harden the lookup.
- The `known` map concurrency issue is worth a follow-up ticket, even if it's low probability today.

## Verdict

**Approve.** Clean implementation of option 1 from the issue. The `PollResult` split is well-structured, `SyncUpdates` is conservative (only touches title/description, preserves status), tests cover the important paths (new vs updated, unchanged, missing ref, empty slice), and the `MarkKnown` seeding prevents false positives on restart. The two "important" findings are pre-existing and don't block this PR.

Issue #3, opened 2026-04-10
Link PRs back to their source work item


jasonmadigan:
Work items that are related should be grouped together in the UI. Currently each item is independent, even when they're part of the same piece of work.

**Examples:**
- Issue #27 spawned PR #30 -- these should be linked
- A PR created by clawdio from a work item should link back to the source
- Multiple dispatch sessions on the same item are already linked (via sessions), but cross-item relationships aren't tracked

**What grouping gives us:**
- Click an issue, see its PRs and their status
- Click a PR, see which issue it addresses
- When reviewing a PR, the context of the original issue is one click away
- The board could show grouped items as stacked cards or with a relationship indicator

**Implementation options:**
1. **Parent-child field on WorkItem** -- a `parent_id` that links a PR work item to its source issue work item
2. **GitHub-derived links** -- parse 'Closes #N' from PR bodies, match source_refs
3. **Manual linking** -- let the user drag one item onto another to link them
4. **Automatic for clawdio branches** -- when push-and-pr creates a PR from a work item's worktree, store the relationship

Option 2 is the most natural for GitHub-sourced items. The data is already there in the PR body (we add 'Closes <source_ref>' in the push-and-pr handler).

PR #44, opened 2026-04-11
Use readable branch names instead of UUIDs

### LINKED ISSUE #7: Use readable branch names instead of UUIDs
Worktree branches are currently named `clawdio/<uuid>` which is ugly and meaningless in git log, PR titles, etc.

Change to `clawdio/<issue-number>-<slugified-title>`, e.g. `clawdio/3-link-prs-to-source-items`.

The slug should be derived from the work item title: lowercase, non-alphanumeric chars replaced with hyphens, truncated to ~50 chars. The issue number comes from the source_ref (e.g. extract `3` from `jasonmadigan/clawdio#3`).

Changes needed:
- `internal/worktree/worktree.go` -- `Create()` needs to accept or derive a readable name
- `internal/session/manager.go` -- pass the readable name when creating worktrees
- Fallback to UUID if no source_ref is available (manual items)

jasonmadigan:
Worktree branches are currently named `clawdio/<uuid>` which is ugly and meaningless in git log, PR titles, etc.

Change to `clawdio/<issue-number>-<slugified-title>`, e.g. `clawdio/3-link-prs-to-source-items`.

The slug should be derived from the work item title: lowercase, non-alphanumeric chars replaced with hyphens, truncated to ~50 chars. The issue number comes from the source_ref (e.g. extract `3` from `jasonmadigan/clawdio#3`).

Changes needed:
- `internal/worktree/worktree.go` -- `Create()` needs to accept or derive a readable name
- `internal/session/manager.go` -- pass the readable name when creating worktrees
- Fallback to UUID if no source_ref is available (manual items)

Closes jasonmadigan/clawdio#7


jasonmadigan (commented):
No interface to worry about -- `Manager` is used directly. The review is complete.

---

Here's the review:

## CI Status

Passing.

## Summary

Replaces UUID-based worktree branch names (`clawdio/<uuid>`) with readable names derived from the source issue (`clawdio/<issue-number>-<slugified-title>`), falling back to UUID when no source ref exists. New `branch.go` handles slug generation, and `Remove` now discovers the actual branch name from the worktree instead of reconstructing it.

## Findings

**Minor:**

1. **`branch.go:49` -- slugify truncates by bytes, not runes.** `len(s)` and `s[:50]` operate on bytes. Since the input has already been lowercased and stripped to `[a-z0-9-]` by that point, this is fine in practice -- those characters are all single-byte UTF-8. Just noting it's safe because of the ordering of operations, not by accident.

2. **`branch_test.go:37` -- long title test case lands on a word boundary.** The expected output `this-is-a-very-long-title-that-should-be-truncated` is exactly 50 chars and happens to end cleanly, so the mid-word cleanup branch (`LastIndex`) isn't exercised. A test where the 50th char falls mid-word would be more thorough, but this is minor.

3. **`worktree.go:179` -- `worktreeBranch` runs `rev-parse` on the worktree path.** If the worktree is in a broken state (e.g. partially removed), this silently returns empty and the branch won't be cleaned up. The old code would have always attempted cleanup. Acceptable tradeoff -- orphaned branches are harmless -- but worth noting.

## Suggestions

- Consider adding one more test case to `TestBranchName` where truncation actually hits mid-word (e.g. a title that produces a slug where char 50 is inside a word), to exercise the `LastIndex` fallback path.

## Verdict

**Approve.** Clean, well-structured change. Good separation of the slug logic into its own file with thorough tests. The `Remove` fix to discover the branch dynamically is the right approach now that branch names aren't predictable from the item ID alone.

Issue #1, opened 2026-04-10
UI for skills available/editing etc.


jasonmadigan:
I'd like some UI to show/edit skills we have in our list too, from our claude config. I plan to have a bunch of skills that agentify what I generally "Do" with these primatives. For example, if I'm assigned a PR to review, I'll review it - I'll look at the code, I'll dispatch some agents to look at the codebase. After review, I'll have some comments to dispatch to certain line numbers, or perhaps the PR is fine and we land an ack on it. I want to encapsulate all the stuff I "do" as skills (feels like the right claude primitive?)

PR #46, opened 2026-04-11
Sync work item status back to GitHub (labels and/or project state)

### LINKED ISSUE #28: Sync work item status back to GitHub (labels and/or project state)
When clawdio works on a GitHub-sourced item, the status should be reflected back to GitHub. This applies to ALL skills, not just implement-issue -- it's an orchestrator-level concern.

**Per skill type:**

| Skill | On dispatch | On completion | On PR merged |
|---|---|---|---|
| implement-issue | Label `clawdio:in-progress`, comment 'agent started' | Label `clawdio:review`, comment with PR link | Remove labels, close issue |
| review-pr | Comment 'review started' | Post review comments, approve/request changes | - |
| address-feedback | Comment 'addressing feedback' | Comment 'changes pushed' | - |
| default | Label `clawdio:in-progress` | Label `clawdio:review` | Remove labels |

**Implementation:**

This should NOT be in the skills themselves -- it's the orchestrator's job. The session manager should:

1. **On dispatch:** check if source is github, run `gh issue edit --add-label` or `gh pr comment`
2. **On completion:** update labels, post PR link comment
3. **On merge:** clean up labels

This keeps skills focused on the actual work and makes status sync consistent regardless of which skill is used.

**Config:**
```json
"github_status_sync": {
  "enabled": true,
  "use_labels": true,
  "post_comments": true
}
```

Labels: `clawdio:in-progress`, `clawdio:review`, `clawdio:done`
Comments: terse, one-line, with links.

jasonmadigan:
When clawdio works on a GitHub-sourced item, the status should be reflected back to GitHub. This applies to ALL skills, not just implement-issue -- it's an orchestrator-level concern.

**Per skill type:**

| Skill | On dispatch | On completion | On PR merged |
|---|---|---|---|
| implement-issue | Label `clawdio:in-progress`, comment 'agent started' | Label `clawdio:review`, comment with PR link | Remove labels, close issue |
| review-pr | Comment 'review started' | Post review comments, approve/request changes | - |
| address-feedback | Comment 'addressing feedback' | Comment 'changes pushed' | - |
| default | Label `clawdio:in-progress` | Label `clawdio:review` | Remove labels |

**Implementation:**

This should NOT be in the skills themselves -- it's the orchestrator's job. The session manager should:

1. **On dispatch:** check if source is github, run `gh issue edit --add-label` or `gh pr comment`
2. **On completion:** update labels, post PR link comment
3. **On merge:** clean up labels

This keeps skills focused on the actual work and makes status sync consistent regardless of which skill is used.

**Config:**
```json
"github_status_sync": {
  "enabled": true,
  "use_labels": true,
  "post_comments": true
}
```

Labels: `clawdio:in-progress`, `clawdio:review`, `clawdio:done`
Comments: terse, one-line, with links.

Closes jasonmadigan/clawdio#28


jasonmadigan (commented):
I've got everything I need. Here's the review:

---

## Review: PR #46 -- Sync work item status back to GitHub

**CI:** Passing.

**Summary:** This is a design spec (not implementation code) for syncing clawdio work item status back to GitHub via labels and comments. It proposes orchestrator-level sync in the session manager and poller, keeping skills clean.

### Findings

**No critical issues.**

**Important:**

1. **`docs/superpowers/specs/` is an odd location.** There's already one other spec in this directory, but the path suggests these are "superpowers" plugin specs. This feature is core clawdio functionality, not a superpowers concern. Consider `docs/specs/` or `docs/design/` instead.

2. **The "out of scope" note about label creation is wrong.** The spec says "gh creates them on first use" -- this isn't accurate. `gh issue edit --add-label` will create the label if it doesn't exist, but with a random colour. Worth noting this is the intended behaviour rather than framing it as "out of scope", or explicitly adding a `gh label create` call on first use to control the colour/description.

3. **Fire-and-forget goroutines need care.** The spec says all sync calls are "fire-and-forget (goroutine, log errors, don't block lifecycle)". This is the right instinct, but the implementation should ensure these goroutines don't outlive the process on shutdown. The manager likely has a shutdown path -- the goroutines should respect context cancellation. Worth a sentence in the spec about graceful shutdown.

4. **`context.Background()` in both dispatch and finish.** The spec shows `go m.syncOnDispatch(context.Background(), ...)`. If these are fire-and-forget, `context.Background()` is fine, but it contradicts the shutdown concern above. Consider deriving from the manager's lifecycle context with a timeout.

**Minor:**

5. **Spec says `SyncStatuses()` but the actual method is split** -- `SyncStatuses` handles open/closed/merged transitions (line 36 of poller.go), but there's a separate `SyncUpdates` for title/description changes. The spec correctly targets `SyncStatuses`, just noting for clarity.

6. **The merged-PR cleanup path only handles label removal.** When a PR is merged for an `implement-issue` item, the spec's table says "Remove labels, close issue" -- but the implementation section only shows `RemoveLabelsWithPrefix`. The issue-closing step isn't detailed anywhere in the integration points section. This might already happen via GitHub's "closes #N" in the PR body, but worth making that assumption explicit.

7. **The `session/sync.go` split seems unnecessary.** The spec lists both `internal/session/sync.go` (sync decision logic) and `internal/github/sync.go` (CLI wrappers). The decision logic is a small switch on skill name -- it could live in the manager methods directly rather than a separate file, avoiding an extra file for ~30 lines of routing.

### Verdict

**Approve.** It's a clean design spec that correctly identifies the integration points and keeps the sync concern at the orchestrator level. The assumptions about existing code all check out. The issues above are minor and can be addressed during implementation.

jasonmadigan:
stop writing the spec now, implement it.

jasonmadigan (commented):
I have a thorough understanding of the changes. Here's my review:

---

## CI Status

## Summary

Adds orchestrator-level GitHub status sync -- labels and comments posted to issues/PRs when clawdio dispatches, completes, or merges work. Config-driven (opt-in), per-skill routing, fire-and-forget with bounded context. Clean separation: sync logic lives in the session manager and poller, not in skills.

## Findings

### Important

1. **`syncOnDispatch` uses `sk.Name` but `syncOnComplete` uses `resolveSkillName(item)` -- inconsistent skill resolution.**

   In `syncOnDispatch` (`manager.go:599`), the skill name comes from `sk.Name` (the `DispatchSkill` param). In `syncOnComplete` (`manager.go:636`), it falls back to `resolveSkillName(item)` which derives from `item.SkillName` then `item.ItemType`. These could diverge if the `DispatchSkill.Name` doesn't match `item.SkillName`. At dispatch time, `resolveSkillName` should work fine too (the item has its skill name set before dispatch), so using it consistently in both places would eliminate this class of bug.

2. **`EnsureLabel` called on every dispatch, not just first use.**

   `syncOnDispatch` (`manager.go:584-588`) calls `EnsureLabel` for all three labels on every single dispatch when labels are enabled. That's 3 extra `gh label create --force` calls per dispatch. The `--force` flag means this is idempotent and won't fail, but it's unnecessary network chatter. A `sync.Once` or a `map[string]bool` of repos where labels have been ensured would be cleaner.

3. **`RemoveLabel` error silently swallowed in `syncOnComplete`.**

   `manager.go:641`: `github.RemoveLabel(ctx, item.SourceRef, "clawdio:in-progress")` -- the error return is ignored. If the label doesn't exist (item was dispatched with labels disabled, then completed with labels enabled), this will log noise from the `gh` CLI but the code won't know. Consistent with the fire-and-forget approach, but should at least log like the other calls do.

### Minor

4. **Poller cleans labels on merge but not on close.** The `SyncStatuses` method removes labels when `state == "merged"` but not when `state == "closed"`. If an issue is closed without merging (e.g. won't-fix), `clawdio:in-progress` labels will stick around. May be intentional (the spec only mentions merge cleanup), but worth calling out.

5. **Test coverage is thin.** Tests cover `resolveSkillName` and invalid-sourceref validation, which is good. But there are no tests for the actual sync routing logic (`syncOnDispatch`, `syncOnComplete`) -- e.g. verifying that `implement-issue` produces label+comment while `review-pr` only produces a comment. The `gh` calls are hard to mock with the current package-level function design, but even a table-driven test that exercises the config guards (`Enabled`, `UseLabels`, `PostComments`) would catch regressions.

6. **Design spec included in the PR.** `docs/specs/2026-04-11-github-status-sync-design.md` is useful documentation, though a couple of its code snippets are now slightly stale vs the implementation (e.g. the spec still shows a separate `session/sync.go` in the file list, which was dropped per review feedback). Minor, but could confuse future readers.

## Suggestions

- Use `resolveSkillName(item)` in `syncOnDispatch` too (or pass skill name as a string to both methods) for consistency.
- Gate `EnsureLabel` behind a per-repo "already ensured" check to avoid 3 redundant calls per dispatch.
- Log the `RemoveLabel` error at `manager.go:641` like the other sync calls.
- Consider also cleaning labels on close, or document why it's merge-only.

## Verdict

The architecture is sound -- sync as an orchestrator concern, fire-and-forget with bounded contexts, clean per-skill routing. The issues above are all minor/moderate. **Approve** with the suggestion to address #1 and #3 as quick follow-ups.

jasonmadigan (commented):
I have everything I need. Here's the review:

---

## CI Status

Pending -- build-and-test is still running. Findings below are based on code review.

## Summary

Adds orchestrator-level GitHub status sync: labels and comments are applied to GitHub issues/PRs when clawdio dispatches, completes, or merges work. Config-driven (opt-in via `github_status_sync`). Sync logic lives in session manager and poller, not in skills -- correct approach per the spec.

## Findings

### Important

1. **Inconsistent struct alignment in constructor** (`internal/session/manager.go:90-91`)

   The `sessions` and `labelsEnsured` fields use wider alignment than the fields above them. Not a bug, but inconsistent with the rest of the struct literal and will trip `gofumpt`/linters.

   ```go
   sessions:      make(map[string]*runningSession),
   labelsEnsured: make(map[string]bool),
   ```

   Should match the shorter alignment used by the other fields above (single-space after colon).

2. **`ctx` variable shadowed in `Dispatch`** (`internal/session/manager.go:142`)

   The `Dispatch` method receives `ctx context.Context` as a parameter, then shadows it:
   ```go
   ctx, cancel := context.WithTimeout(m.ctx, 30*time.Second)
   ```
   This isn't a bug because the new `ctx` is only used inside the goroutine, and the original `ctx` isn't used after this point. But it's a shadow that could trip `go vet -shadow` or confuse future readers. Consider naming it `syncCtx`.

3. **`RemoveLabelsWithPrefix` is not fire-and-forget in poller** (`internal/poller/poller.go:74-77`)

   The design spec says sync calls should be fire-and-forget (goroutine, don't block lifecycle). But in the poller, `RemoveLabelsWithPrefix` is called synchronously -- it blocks the status sync loop. If the `gh` call hangs or is slow, it delays processing of other items. Consider wrapping in a goroutine or at least noting that this is intentional (poller runs periodically so blocking is acceptable).

### Minor

4. **`EnsureLabel` uses `--force` flag** (`internal/github/sync.go:84`)

   Using `--force` with `gh label create` overwrites existing label colour/description each time. If a user has customised the label colours in their repo, clawdio will reset them on first dispatch. The spec said "if it fails (label already exists), that's fine -- carry on." Consider dropping `--force` and ignoring the error instead, which preserves any user customisation.

5. **Design spec in the repo** (`docs/specs/2026-04-11-github-status-sync-design.md`)

   Good to have the design doc, but it duplicates the PR description almost verbatim. If this is meant as durable documentation, fine. Just noting it's another thing to keep in sync.

6. **`PostComment` error for trailing `#`** (`internal/github/sync_test.go:30`)

   `ParseSourceRef("trailing#")` returns `("trailing", "")` -- empty number, caught correctly. Good edge case test.

## Suggestions

- Extract the label swap pattern (remove old + add new) into a helper like `SwapLabel(ctx, ref, from, to)` since it appears in both `syncOnComplete` cases (`implement-issue` and `default`). Reduces duplication and makes the swap atomic-ish in intent.

- The `labelsEnsured` map grows unboundedly (one entry per repo). Not a real problem at current scale, but worth noting if clawdio ever monitors many repos.

## Verdict

The implementation matches the spec well. The architecture is sound -- sync at the orchestrator level, skills stay clean. Code is clear and well-tested for the pure logic (`resolveSkillName`). The gh CLI wrapper tests sensibly cover the invalid-ref edge cases without calling out to real GitHub.

**Request changes** for the alignment issue (#1) -- it's the kind of thing that'll annoy linters. The rest are suggestions, not blockers.

jasonmadigan (commented):
Right -- deliberately not using `--force` to preserve user customisations. That's a good design choice, matching what the spec says.

Here's my review:

---

## CI Status

**Passing.** `build-and-test` green.

## Summary

Adds orchestrator-level GitHub status sync -- when clawdio dispatches, completes, or merges work on GitHub-sourced items, it reflects that status back via labels (`clawdio:in-progress`, `clawdio:review`, `clawdio:done`) and comments. The sync lives in the session manager and poller, not in individual skills. Configurable via `github_status_sync` in config.

## Findings

### Important

1. **`RemoveLabelsWithPrefix` in poller is synchronous/blocking** (`sync.go:44-60`) -- The poller calls `RemoveLabelsWithPrefix` inline during `SyncStatuses`, which makes N+1 `gh` CLI calls (1 to list labels, then 1 per matching label). The dispatch and complete paths correctly fire sync in goroutines, but the poller blocks on this. For a single item this is fine, but if many items close at once during a poll cycle, the poller will stall. Not critical now, but worth noting as a potential bottleneck.

2. **No sync on session error** (`manager.go:543-551`) -- `syncOnComplete` only fires when `status == model.SessionCompleted`. If a session errors out, the `clawdio:in-progress` label stays on the issue permanently. Consider handling `SessionErrored` too -- at minimum removing the in-progress label, possibly posting a comment noting the failure.

### Minor

3. **Whitespace-only reformatting** (`manager.go:162`, `manager.go:313-314`) -- A couple of alignment changes in `SendMessage` and `run()` that aren't related to this feature. Harmless but adds noise to the diff.

4. **Design spec checked in** (`docs/specs/2026-04-11-github-status-sync-design.md`) -- 162-line spec document. No issue with having it, but worth confirming this is intentional as a project convention vs one-off.

## Suggestions

- For finding #2, add a `syncOnError` path (or widen the condition in `finish()`) that removes the in-progress label when a session errors. Something like:

```go
if status == model.SessionCompleted || status == model.SessionErrored {
    // ...
    if status == model.SessionCompleted {
        m.syncOnComplete(syncCtx, item)
    } else {
        // remove in-progress label on failure
        github.RemoveLabel(syncCtx, item.SourceRef, "clawdio:in-progress")
    }
}
```

- Consider wrapping the poller's `RemoveLabelsWithPrefix` call in a goroutine to match the pattern used in the session manager, so the poll loop doesn't block on slow GitHub API calls.

## Verdict

**Approve** -- this is clean, well-structured work. The sync logic is correctly placed at the orchestrator level, config is sensible, the label-ensure-once pattern avoids redundant calls, and tests cover the right things. Finding #2 (stale labels on error) is a real gap but could be a follow-up.

Issue #12, opened 2026-04-10
Drop-in should continue the agent's session, not start fresh


jasonmadigan:
The 'copy cd + claude' button currently gives `cd <worktree> && claude` which starts a brand new session with no context from the agent's work.

It should use `claude --continue` or `claude --resume <session-id>` so the user picks up where the agent left off, with full conversation history and context.

The agent's session ID is available from our database. Two options:
- `cd <worktree> && claude --continue` -- continues the most recent session in that directory (simplest, should work if session was persisted)
- `cd <worktree> && claude --resume <session-id>` -- explicitly resumes by ID (more reliable)

Need to verify that sessions spawned with `--print` are persisted and resumable. If `--no-session-persistence` is not set, they should be.

Issue #10, opened 2026-04-10
URL routing for work items and views


jasonmadigan:
The UI is a single-page app with no URL routing. Everything lives at `/`. This means:

- Can't link to a specific work item (`/items/<id>`)
- Can't bookmark a view (`/board`, `/skills`, `/settings`)
- Browser back/forward doesn't work
- Refreshing always goes to the board

Add client-side routing so URLs reflect the current view:
- `/` or `/board` -- kanban board
- `/board/<item-id>` -- board with item panel open
- `/skills` -- skills view
- `/settings` -- settings view

Use the History API (pushState/popState) or a lightweight router like wouter.

Issue #9, opened 2026-04-10
Bidirectional agent sessions: answer questions and approve tools


jasonmadigan:
Currently sessions run with `--permission-mode bypassPermissions` (one-shot, fire and forget).

**What exists already (but isn't wired up):**
- Model types: `ControlRequest`, `AgentState` (awaiting_input, awaiting_approval)
- API endpoints: `/api/sessions/{id}/message`, `approve`, `deny`, `cancel`
- React Panel: approval box UI, input field for answering questions
- Session manager: `SendMessage()`, `ApproveToolUse()`, `DenyToolUse()` methods

**What's missing:**
- Actually using `--input-format stream-json` (caused output buffering issues in testing)
- Switching to `--permission-mode default` so the agent asks for permissions
- Testing the full bidirectional flow end-to-end
- The stdin close fix needs to be reverted when this is enabled

**Blocker from initial testing:** `--input-format stream-json` may change how claude buffers output. Needs investigation.

**When this lands:**
- Remove the non-interactive constraints from all dispatch skills (`skills/*.md`). Currently all skills say 'Do NOT wait for user input or approval' and 'Do NOT use planning skills that require approval steps'. These exist because sessions are one-shot. With bidirectional support, agents could use planning/approval workflows interactively.
- Update skill docs to note which skills benefit from interactive mode vs fire-and-forget.

## Permission mode toggle

Currently hardcoded to `--permission-mode bypassPermissions` in manager.go. Needs to be configurable at three levels:

**Global default (config.json):**
```json
"permission_mode": "default"
```
Options: `default` (agent asks for permission), `bypassPermissions` (agent runs autonomously). Default should be `default` for safety.

**UI toggle:**
Prominent toggle in the header (like the autopilot button). Visually obvious when bypass is on -- power-user mode.

**Per-dispatch override:**
Permission mode dropdown in the dispatch modal. Override the global setting for a single run.

When `default` mode is active, bidirectional comms are essential -- the agent asks permission and the UI handles approve/deny. When `bypassPermissions` is on, that UI is dormant.

## Prerequisites (now met)

- #85 (state machine) -- merged
- #86 (adapters) -- merged
- #87 (contracts) -- merged

jasonmadigan:
**Scope expansion: permission mode toggle**

Currently hardcoded to \`--permission-mode bypassPermissions\` in manager.go. This needs to be configurable at three levels:

**Global default (config.json):**
\`\`\`json
[...]

PR #48, opened 2026-04-11
Code quality: fix bugs, dead code, duplication, and structural issues

### LINKED ISSUE #47: Code quality: fix bugs, dead code, duplication, and structural issues
Full codebase review surfaced these issues, prioritised:

## Bugs / dead code (fix first)

1. **stdin bug in manager.go** -- stdin pipe is created then immediately closed (line 268), but `SendMessage`, `ApproveToolUse`, `DenyToolUse` all try to write to it. Either a crash waiting to happen or dead code masquerading as features. Fix: remove the close if bidirectional is planned, or remove the write methods.

2. **Dead code to delete:**
   - `ManualAdapter` in `adapter/manual.go` -- never instantiated
   - `WSMessage` and `SessionOutput` in `model/model.go` -- never referenced
   - `BroadcastSessionOutput` in `server.go` -- superseded by `BroadcastAgentEvent`

3. **Livereload watcher never closes** -- `WatchWebDir` creates an fsnotify.Watcher with no shutdown path. Pass a context or store on Server for cleanup.

## Structural (split / extract)

4. **`handlers.go` is 660+ lines** -- split by resource:
   - `work_item_handlers.go` -- CRUD, dispatch, push/PR, review
   - `skill_handlers.go` -- skill CRUD, dirs, mapping
   - `session_handlers.go` -- message, approve, deny, cancel

5. **`handleAssistantMessage` is 100+ lines** -- extract:
   - `formatToolDetail(name, inputJSON)` for tool display formatting
   - A dedup struct that tracks emitted lengths with an `emitNewContent()` method

6. **`run()` is 145 lines** -- extract:
   - `buildCommand(sess) (*exec.Cmd, io.ReadCloser, error)` for arg/cmd/pipe setup
   - Stream parsing loop as its own method

7. **`main.go` onStatus closure** -- 20-line closure capturing 5 variables. Should be a method on a struct with access to store and action executor.

## Duplication

8. **IsPR check in 3 places** -- `handlers.go:217`, `manager.go:90`, `actions.go:93` all independently check `item.ItemType == TypePRReview || ...`. Add `func (w WorkItem) IsPR() bool` to model.

9. **ParseSourceRef reimplemented** -- `adapter/github.go` and `worktree/branch.go` both split on `#` to parse `owner/repo#42`. Use `github.ParseSourceRef` everywhere.

10. **ProcessPostSession pattern** -- both post-session methods repeat: marshal content, evaluate autonomy, optionally execute, create action. Extract a shared helper.

## Type safety

11. **ActionMode type is dead** -- model defines `ActionMode` with constants but `Engine.Evaluate` returns bare strings. All callers compare against string literals. Either use the type or delete it.

## Infrastructure

12. **Hand-rolled frontmatter parser** -- 89 lines of bespoke YAML parsing. Fragile assumptions about whitespace. Use `gopkg.in/yaml.v3` instead.

13. **Migration strategy won't scale** -- bare SQL strings, no ordering, no version tracking. Errors detected by string matching (`duplicate column`). A `schema_version` table with sequential numbering is the standard fix.

14. **Error handling inconsistency** -- `pollNotifications` silently swallows errors (`continue`) but `pollIssues`/`pollPRs` return immediately. Should be consistent.

## Naming / comments

15. **Obvious comments** -- strip comments like `// Cancel stops a running session`, `// Close closes the store`. Method names already say this.

16. **ResolveRepo silent magic** -- probes `~/Work`, `~/work`, `~/src`, `~/Projects` without logging. At minimum log when a guess succeeds.

jasonmadigan:
Full codebase review surfaced these issues, prioritised:

## Bugs / dead code (fix first)

1. **stdin bug in manager.go** -- stdin pipe is created then immediately closed (line 268), but `SendMessage`, `ApproveToolUse`, `DenyToolUse` all try to write to it. Either a crash waiting to happen or dead code masquerading as features. Fix: remove the close if bidirectional is planned, or remove the write methods.

2. **Dead code to delete:**
   - `ManualAdapter` in `adapter/manual.go` -- never instantiated
   - `WSMessage` and `SessionOutput` in `model/model.go` -- never referenced
   - `BroadcastSessionOutput` in `server.go` -- superseded by `BroadcastAgentEvent`

3. **Livereload watcher never closes** -- `WatchWebDir` creates an fsnotify.Watcher with no shutdown path. Pass a context or store on Server for cleanup.

## Structural (split / extract)

4. **`handlers.go` is 660+ lines** -- split by resource:
   - `work_item_handlers.go` -- CRUD, dispatch, push/PR, review
   - `skill_handlers.go` -- skill CRUD, dirs, mapping
   - `session_handlers.go` -- message, approve, deny, cancel

5. **`handleAssistantMessage` is 100+ lines** -- extract:
   - `formatToolDetail(name, inputJSON)` for tool display formatting
   - A dedup struct that tracks emitted lengths with an `emitNewContent()` method

6. **`run()` is 145 lines** -- extract:
   - `buildCommand(sess) (*exec.Cmd, io.ReadCloser, error)` for arg/cmd/pipe setup
   - Stream parsing loop as its own method

7. **`main.go` onStatus closure** -- 20-line closure capturing 5 variables. Should be a method on a struct with access to store and action executor.

## Duplication

8. **IsPR check in 3 places** -- `handlers.go:217`, `manager.go:90`, `actions.go:93` all independently check `item.ItemType == TypePRReview || ...`. Add `func (w WorkItem) IsPR() bool` to model.

9. **ParseSourceRef reimplemented** -- `adapter/github.go` and `worktree/branch.go` both split on `#` to parse `owner/repo#42`. Use `github.ParseSourceRef` everywhere.

10. **ProcessPostSession pattern** -- both post-session methods repeat: marshal content, evaluate autonomy, optionally execute, create action. Extract a shared helper.

## Type safety

11. **ActionMode type is dead** -- model defines `ActionMode` with constants but `Engine.Evaluate` returns bare strings. All callers compare against string literals. Either use the type or delete it.

## Infrastructure

12. **Hand-rolled frontmatter parser** -- 89 lines of bespoke YAML parsing. Fragile assumptions about whitespace. Use `gopkg.in/yaml.v3` instead.

13. **Migration strategy won't scale** -- bare SQL strings, no ordering, no version tracking. Errors detected by string matching (`duplicate column`). A `schema_version` table with sequential numbering is the standard fix.

14. **Error handling inconsistency** -- `pollNotifications` silently swallows errors (`continue`) but `pollIssues`/`pollPRs` return immediately. Should be consistent.

## Naming / comments

15. **Obvious comments** -- strip comments like `// Cancel stops a running session`, `// Close closes the store`. Method names already say this.

16. **ResolveRepo silent magic** -- probes `~/Work`, `~/work`, `~/src`, `~/Projects` without logging. At minimum log when a guess succeeds.

Closes jasonmadigan/clawdio#47


jasonmadigan (commented):
Right. `config.go:66` could use `ParseSourceRef` -- it only needs `parts[0]` which is the repo, and it ignores the number. That's a valid dedup target. Minor point though.

Now let me write up the review.

---

## CI Status

**Passing.** `build-and-test` green.

## Summary

Addresses a 16-item code quality audit: fixes the stdin pipe bug in session manager, removes dead code (`ManualAdapter`, `WSMessage`, `SessionOutput`, `BroadcastSessionOutput`, `ActionMode`), adds context-based shutdown for the livereload watcher, splits `handlers.go` into three files by resource, adds `WorkItem.IsPR()` to deduplicate type checks, uses `github.ParseSourceRef` in the GitHub adapter, fixes silent error swallowing in `pollNotifications`, adds logging to `ResolveRepo`, and strips obvious comments.

+239 / -280 across 11 files. Net negative diff -- good.

## Findings

### Important

1. **`config.go:66` still splits `sourceRef` by hand.** The issue (#9) called out `adapter/github.go` and `worktree/branch.go` as duplicating `ParseSourceRef`. The adapter was fixed, but `config.go` still does `strings.SplitN(sourceRef, "#", 2)` and takes `parts[0]`. It could use `repo, _ := github.ParseSourceRef(sourceRef)` for consistency. Minor -- it works fine as-is, but it's the same pattern the PR set out to centralise.

2. **Several audit items explicitly deferred.** The PR description lists 16 items but only addresses roughly 8-9 of them. Not addressed:
   - #5: `handleAssistantMessage` extraction (formatToolDetail, dedup struct)
   - #6: `run()` extraction (buildCommand helper)
   - #7: `main.go` onStatus closure (still a 19-line closure capturing multiple variables)
   - #10: ProcessPostSession duplication
   - #12: Hand-rolled frontmatter parser (still 88 lines of bespoke YAML)
   - #13: Migration strategy (still uses `duplicate column` string matching)

   These are all called out in the PR body but not implemented. If this is intentional scoping (fix the bugs/dead code now, defer structural work), that's fine -- but the PR body reads as if everything was done.

### Minor

3. **`handleCheckPR` moved but not split.** It now lives in `work_item_handlers.go` alongside the other work-item handlers, which is the right place. No issue here, just noting the PR added it inline rather than leaving it where it was in the diff -- cleaner result.

4. **Renamed file carries some git noise.** `handlers.go` -> `work_item_handlers.go` is a rename, so the diff is large but the actual content changes are small (IsPR usage, comment removal, function reordering). The split is clean.

## Suggestions

- Update the PR description to clarify which items are addressed and which are deferred to a follow-up. Right now the description lists all 16 as if they're all done.
- `config.go:66`: one-line fix to use `ParseSourceRef` for consistency.

## Verdict

**Approve.** The changes that _are_ here are solid: the stdin fix is correct (pipe stays open for bidirectional use), dead code removal is clean, handler split is well-scoped, `IsPR()` dedup is the right pattern, livereload context cancellation is proper. The PR description just oversells the scope -- about half the listed items are deferred. No bugs introduced.

jasonmadigan:
Do all of those deferrals, we need to complete the work. stop being lazy.

jasonmadigan:
I mean it. Do all the deferred work, now.

jasonmadigan (commented):
All changes are in two commits. Here's my review:

---

## CI status

**Pending** -- `build-and-test` is still running. No failures yet.

## Summary

Addresses ~9 of the 16 items from #47: fixes the stdin pipe bug, removes dead code (ManualAdapter, WSMessage, SessionOutput, BroadcastSessionOutput, ActionMode), adds context-based shutdown for the livereload watcher, splits the monolithic handlers.go into three files by resource, consolidates duplicated IsPR checks into a model method, replaces inline sourceRef parsing with `github.ParseSourceRef`, makes notification poll errors visible, strips obvious comments, and logs ResolveRepo guesses.

## Findings

### Important

1. **Items 5-7, 10, 12-13 not addressed** -- The structural extractions (handleAssistantMessage, run(), onStatus closure), ProcessPostSession deduplication, frontmatter parser replacement, and migration strategy improvements are all deferred. This is fine for a single PR -- just noting they're still open. Worth a follow-up issue or keeping #47 open for the remaining items.

### Minor

2. **`IsPR()` defined before its receiver type** (`model.go:25-27`, before `WorkItem` struct at line 29). Valid Go, but unconventional -- most codebases define methods after the struct. Consider moving it below the struct definition for readability.

3. **`writeJSON` lives in `session_handlers.go`** but is used by all three handler files. It's a package-level utility, so putting it in `session_handlers.go` is slightly arbitrary. Not a problem, but a `helpers.go` or leaving it in the work_item file (the largest consumer) might be more discoverable. Very minor.

4. **Notification poll error change** (`adapter/github.go:142`) -- switching from `continue` to `return error` is correct. Verified the caller in the poller logs and skips the cycle, so no crash risk. Good change.

## Verdict

**Do Not Approve** (assuming CI passes). Clean, net-negative-lines cleanup that fixes real bugs (stdin pipe, watcher leak) and improves structure without changing behaviour. The remaining items need to be completed before I'll even consider approving this. No more slacking.

jasonmadigan (commented):
The worktree this session was running in got detached by the subagent's checkout. I have all the information I need from the full diff I already read. Let me write the review based on my thorough analysis.

Here's my review of PR #48:

---

## CI Status

No checks configured on this branch.

## Summary

Large cleanup PR addressing 16 items from a codebase audit. Covers dead code removal, handler file splitting, dedup/stream parsing extraction, `IsPR()` helper, `ParseSourceRef` reuse, yaml.v3 for frontmatter, schema_version migration tracking, and various comment/naming fixes. Net -18 lines across 18 files.

## Findings

### Critical

**1. stdin bug fix is incomplete -- stderrBuf access is now unsafe**

`buildCommand` creates a local `stderrBuf` and sets `cmd.Stderr = &stderrBuf`. But `stderrBuf` is local to `buildCommand`, so after it returns, `run()` accesses the buffer via `cmd.Stderr.(*bytes.Buffer)` type assertion. This works because the `*bytes.Buffer` pointer survives, but it's fragile and non-obvious -- the variable's lifetime is only accidentally correct because `cmd` holds the pointer. A future refactor could easily break this. Consider returning the buffer alongside the cmd, or storing it on `runningSession`.

**2. stdin fix keeps pipe open but has no cleanup**

The PR correctly fixes the original bug (stdin was closed immediately, then write methods tried to use it). stdin is now stored in `runningSession.stdin`. However, stdin is never explicitly closed when the session ends. The `defer` block in `run()` only deletes from the map and calls `cancel()`. If the process exits, the OS reclaims the pipe, but it's still worth closing explicitly for correctness, e.g. in the defer block or after `cmd.Wait()`.

### Important

**3. Migration strategy has an edge case for existing databases**

`runMigrations` does `SELECT COUNT(*) FROM schema_version`, but existing databases created before this PR won't have a `schema_version` table. The schema DDL uses `CREATE TABLE IF NOT EXISTS`, so on a fresh `New()` call the table exists. But on an existing database that already has data and was created with the old code, `schema_version` won't exist and the SELECT will fail. The schema DDL runs before migrations in `New()`, so this should be fine in practice -- but confirm that the full schema DDL (including the new `schema_version` table) runs every time, not just on first init.

**4. `evaluateAndCreate` semantics differ from `ProcessPostSession`**

The new `evaluateAndCreate` helper skips creating the action record entirely when mode is `"blocked"`. But `ProcessPostSession` (git push/PR) still creates action records unconditionally -- push actions are always recorded even when blocked. This asymmetry is documented with a comment ("git push is always created ... for backwards compat"), which is good, but the helper's behaviour should be documented too -- callers need to know that `"blocked"` means no record at all.

**5. No test coverage for `evaluateAndCreate` or `formatToolDetail`**

Both are newly extracted functions with non-trivial logic. `formatToolDetail` has branching on tool names and input parsing. `evaluateAndCreate` has three paths (autonomous, blocked, pending). Neither has tests.

### Minor

**6. `writeJSON` defined in `session_handlers.go` but used across all handler files**

It's a shared utility. Having it in one of the split files works but is arbitrary. Not blocking.

**7. `handleCheckPR` moved but also has a new boundary between the existing handlers**

The handler existed before but got relocated during the split. It's now between `handleMergePR` and `handleListPendingActions`. Fine, just noting for anyone reading the diff.

**8. Frontmatter test relaxation**

`TestParseFrontmatterEmptyPostActions` previously asserted `PostActions != nil` (empty slice, not nil). The yaml.v3 decoder returns nil for an empty sequence. The test was updated to only check `len() == 0`, which is correct for Go usage (nil and empty slice behave identically with `len`), but worth noting the semantic change.

## Suggestions

1. Return `*bytes.Buffer` from `buildCommand` alongside the cmd/stdin/stdout, so the ownership is explicit rather than relying on pointer survival through `cmd.Stderr`.
2. Add `stdin.Close()` in the defer block of `run()`, or after `cmd.Wait()` returns.
3. Add a brief test for `formatToolDetail` -- it's a pure function, easy to test.
4. Consider wrapping each migration step in a transaction (the current approach does `db.Exec` without a tx, so a failure mid-migration leaves partial state).

## Verdict

**Request changes.** The core refactoring is solid and well-structured. The stdin fix addresses a real bug, and the handler splitting and dedup extraction meaningfully improve readability. However, the stderrBuf lifetime issue (#1) and missing stdin cleanup (#2) should be addressed before merging. The migration edge case (#3) should be confirmed as safe. Everything else is minor or nice-to-have.

jasonmadigan (commented):
Good, only one definition. I've now reviewed the full diff. Here's the review:

---

## CI Status

No CI checks configured on this branch.

## Summary

Addresses 16 items from a codebase audit (jasonmadigan/clawdio#47). Fixes the stdin pipe bug, removes dead code, splits a 660-line handler file, introduces proper migration versioning, replaces a hand-rolled YAML parser, extracts helper functions, and removes duplication across the board.

## Findings

### Important

1. **`evaluateAndCreate` executes before checking `blocked`** -- `internal/session/actions.go:97-108`. The ordering is: check `autonomous` -> exec -> check `blocked` -> create record. Since `Evaluate` returns a single string and these are mutually exclusive branches, this is logically correct. But structurally it reads as "we might execute, then check if we should have". A guard `if mode == "blocked" { return nil }` at the top would be clearer and future-proof against someone adding a mode that's both "execute something" and "don't create a record". Minor style issue, not a bug.

2. **`handleCheckPR` moved but not split** -- `internal/server/work_item_handlers.go:359-395`. The PR description says handlers were split into work_item/skill/session files (item 4), and session + skill handlers were moved. But `handleCheckPR` was moved *within* `work_item_handlers.go` to a different position in the file. This is fine semantically but the diff makes it look like the function was recreated rather than just repositioned. Git sees it as delete + add rather than a move. No action needed, just noting for the review record.

### Minor

3. **`writeJSON` ignores `Encode` error** -- `internal/server/helpers.go:11`. `json.NewEncoder(w).Encode(v)` can fail (e.g. a type that doesn't marshal). Since `WriteHeader` is already called, you can't change the status code, but logging would catch bugs. Very minor since all current callers pass known-good types.

4. **Missing `TypePRFeedback` in some contexts** -- The `IsPR()` method at `internal/model/model.go:42` correctly includes `TypePRFeedback`. Worth confirming that all the old inline checks it replaces also included `TypePRFeedback` -- looking at the diff, the old `actions.go:93` only checked `TypePRReview || TypePRAuthored`, not `TypePRFeedback`. If that was intentional (feedback items shouldn't trigger post-review actions), `IsPR()` may now match *more* items than before in `actions.go`. That said, `actions.go` no longer uses `IsPR()` directly, and the callers in `handlers.go` and `manager.go` did already include `TypePRFeedback`. So this is consistent with the handlers/manager, which was the intent.

5. **Frontmatter test relaxed** -- `internal/skill/frontmatter_test.go:90`. The test for empty `post_actions:` previously asserted `fm.PostActions != nil` (i.e. an initialised empty slice). With `yaml.v3`, this now returns `nil` instead of `[]string{}`. The test was updated to remove the nil check, which is fine -- callers should handle nil and empty slices the same way. Just noting the behavioural change.

## Suggestions

- Consider moving the `blocked` guard in `evaluateAndCreate` to the top of the function, before the `autonomous` check. It's correct either way, but "bail early on blocked" is more defensive.
- Items 5 (`handleAssistantMessage` extraction), 6 (`run()` extraction), and 10 (`evaluateAndCreate`) from the issue are all done well. The `textDedup` struct and `formatToolDetail` function are clean extractions with good test coverage.
- The migration strategy (item 13) is solid. The seed-on-fresh-db approach handles the upgrade path correctly.

## Verdict

**Approve.** This is a clean, well-structured cleanup. The stdin fix is correct (pipe stays open, stored on `runningSession`, closed in the defer). Dead code removal is accurate. The handler split is mechanical and correct. The YAML parser swap and migration versioning are genuine improvements. No regressions, good test additions.

PR #50, opened 2026-04-11
Link PRs back to their source work item

### LINKED ISSUE #3: Link PRs back to their source work item

jasonmadigan:
Work items that are related should be grouped together in the UI. Currently each item is independent, even when they're part of the same piece of work.

**Examples:**
- Issue #27 spawned PR #30 -- these should be linked
- A PR created by clawdio from a work item should link back to the source
- Multiple dispatch sessions on the same item are already linked (via sessions), but cross-item relationships aren't tracked

**What grouping gives us:**
- Click an issue, see its PRs and their status
- Click a PR, see which issue it addresses
- When reviewing a PR, the context of the original issue is one click away
- The board could show grouped items as stacked cards or with a relationship indicator

**Implementation options:**
1. **Parent-child field on WorkItem** -- a `parent_id` that links a PR work item to its source issue work item
2. **GitHub-derived links** -- parse 'Closes #N' from PR bodies, match source_refs
3. **Manual linking** -- let the user drag one item onto another to link them
4. **Automatic for clawdio branches** -- when push-and-pr creates a PR from a work item's worktree, store the relationship

Option 2 is the most natural for GitHub-sourced items. The data is already there in the PR body (we add 'Closes <source_ref>' in the push-and-pr handler).

Closes jasonmadigan/clawdio#3


jasonmadigan (commented):
Good, I have a clear picture. Here's the review:

---

## CI Status

**Passing.** `build-and-test` green (38s).

## Summary

Adds parent-child linking between work items so PRs that close issues are associated in the UI. Parses GitHub closing keywords (`Closes #N`, `Fixes #N`, `Resolves #N`) from PR bodies, stores a `parent_id` on the work item, and surfaces the relationship on both board cards and the detail panel.

## Findings

### Important

**1. Duplicated linking logic between `main.go` and `poller.go`**

The parent-linking logic in `cmd/clawdio/main.go:168-180` and `internal/poller/poller.go:tryLinkParent` are nearly identical. If the linking logic needs to change (e.g. linking multiple parents, different ref formats), both need updating independently. Consider extracting a shared helper, or having the initial import path also call `tryLinkParent` via the poller.

**2. `O(n)` scans on every card render**

`Card.tsx` computes `children` and `parent` by filtering/finding across `allItems` on every render. With a small number of work items this is fine, but it's worth noting the pattern -- if item counts grow, a lookup map built once higher up would be cleaner. Not blocking, but worth a note for future.

### Minor

**3. No `key` accessibility concern on linked items in Panel**

The `linked-item` divs in `Panel.tsx:432-445` use `onClick` but lack `role="button"`, `tabIndex`, and `onKeyDown` handlers. They're not keyboard-accessible. Minor since this is an internal tool, but worth a follow-up.

**4. Regex doesn't handle `\r\n` line endings**

`closingRefRe` uses `\s+` between the keyword and ref, which handles `\n` fine but could be worth verifying against Windows-style line endings in PR bodies fetched from GitHub. In practice GitHub normalises these, so low risk.

**5. `break` after first parent match is intentional but undocumented**

In `main.go:178` -- `break // one parent is enough`. The comment helps, but `tryLinkParent` in `poller.go:152` uses `return` for the same purpose with no equivalent comment. Minor consistency thing.

## Suggestions

- Extract the link-parent-to-child logic into a shared function (e.g. on the store, or a dedicated `linking` package) to avoid the duplication between initial import and poll-based re-check.
- Consider building a `parentMap`/`childMap` at the `Board` or `App` level and passing pre-computed relationships down, rather than filtering `allItems` in each `Card`.

## Verdict

**Approve.** Clean implementation of a useful feature. The regex parsing is well-tested, the migration is additive and safe, the UI changes are minimal and focused. The duplication between `main.go` and `poller.go` is the main thing worth addressing, but it's not blocking -- could be a follow-up.

PR #51, opened 2026-04-11
Global prompt rules for dispatched sessions

### LINKED ISSUE #6: Global prompt rules for dispatched sessions
Need a way to define global rules that apply to all dispatched agent sessions, regardless of which skill is used. Examples:

- Never add Co-Authored-By lines to commits
- Always use British English
- Comment style: terse, lowercase, only where needed

**Current state:** With `--bare` off (the default), dispatched sessions pick up the project's CLAUDE.md naturally. This covers repo-specific rules. But there's no way to set rules that apply across all repos.

**Remaining work:**
- A `global_prompt` field in config.json that gets prepended to every dispatch prompt
- Or a `~/.clawdio/rules.md` file that's always included
- The skill prompts already include 'Do NOT add Co-Authored-By' but this should be global, not per-skill

jasonmadigan:
Need a way to define global rules that apply to all dispatched agent sessions, regardless of which skill is used. Examples:

- Never add Co-Authored-By lines to commits
- Always use British English
- Comment style: terse, lowercase, only where needed

**Current state:** With `--bare` off (the default), dispatched sessions pick up the project's CLAUDE.md naturally. This covers repo-specific rules. But there's no way to set rules that apply across all repos.

**Remaining work:**
- A `global_prompt` field in config.json that gets prepended to every dispatch prompt
- Or a `~/.clawdio/rules.md` file that's always included
- The skill prompts already include 'Do NOT add Co-Authored-By' but this should be global, not per-skill

Closes jasonmadigan/clawdio#6


jasonmadigan (commented):
Here's the review:

---

## CI Status

## Summary

Adds a `global_prompt` config field that gets prepended to every dispatched session's prompt, replacing per-skill duplication of cross-repo rules (like the Co-Authored-By line). Clean, minimal change across 11 files.

## Findings

**Minor:**

1. **Prompt ordering with profile**: `global_prompt` goes first, then skill, then task, then profile (`Additional Instructions`). That ordering makes sense -- global rules > skill instructions > task context > profile additions. Just flagging it's intentional.

2. **No validation on `global_prompt` content**: The field accepts any string, which is fine -- it's a power-user config. No issue, just noting there's no length limit or format enforcement, which is the right call for this.

## Suggestions

None. The implementation is straightforward:

- Config field added cleanly to the struct and example config
- `BuildPrompt` signature extended with the new parameter, empty string is a no-op
- Separator only emitted when global prompt is non-empty (tested)
- Ordering test verifies global > skill > task
- Duplicate "Co-Authored-By" lines removed from all four skill files
- Documentation updated with example and rationale
- All existing tests updated for the new signature

The test coverage is good -- tests the ordering, tests the empty case, and existing tests pass with the new parameter.

## Verdict

**Approve.** This is a clean, well-scoped change that does exactly what the issue asked for. No bugs, no missing edge cases, good test coverage.

PR #54, opened 2026-04-12
Surface dispatch and action errors in the UI

### LINKED ISSUE #49: Surface dispatch and action errors in the UI
When dispatch or post-actions fail (e.g. worktree branch conflict, push error, review post failure), the UI shows nothing -- the button just does nothing or silently fails. The user has no idea what went wrong.

**Current behaviour:**
- 'Start agent' button does nothing if dispatch returns an error
- Post-action buttons sometimes show errors, sometimes don't
- No toast/notification system for transient errors

**Fix:**
- Dispatch modal should show the error if the API returns non-200
- Post-action buttons already have some error display but it's inconsistent
- Consider a toast/notification component for transient errors
- All API calls in the frontend should handle errors visibly

Related: the dispatch failure is often a stale branch (`branch already exists`) which the worktree manager should handle by deleting and retrying. But even when the backend can't recover, the UI must show the error.

jasonmadigan:
When dispatch or post-actions fail (e.g. worktree branch conflict, push error, review post failure), the UI shows nothing -- the button just does nothing or silently fails. The user has no idea what went wrong.

**Current behaviour:**
- 'Start agent' button does nothing if dispatch returns an error
- Post-action buttons sometimes show errors, sometimes don't
- No toast/notification system for transient errors

**Fix:**
- Dispatch modal should show the error if the API returns non-200
- Post-action buttons already have some error display but it's inconsistent
- Consider a toast/notification component for transient errors
- All API calls in the frontend should handle errors visibly

Related: the dispatch failure is often a stale branch (`branch already exists`) which the worktree manager should handle by deleting and retrying. But even when the backend can't recover, the UI must show the error.

Closes jasonmadigan/clawdio#49


jasonmadigan:
## CI Status

Passing (build-and-test, 36s).

## Summary

Adds consistent error surfacing across both backend and frontend. Backend switches from `http.Error` (plain text) to a `writeError` helper that returns JSON `{"error": "..."}`. Frontend gets a toast notification system for transient errors, inline error display in modals, and the `api.ts` request helper now extracts error messages from JSON response bodies. The worktree manager also gains a `cleanStaleBranch` recovery path for the common "branch already exists / already checked out" failure.

## Findings

### Important

1. **`cleanStaleBranch` does `git branch -D` unconditionally** (`worktree.go:102-105`). After pruning/removing the stale worktree, the branch is force-deleted so `-b` can recreate it. This means any unpushed commits on that branch are silently lost. The recovery path in `Create` then does `git worktree add wtPath branchName` (line 50) -- but the branch was just deleted on line 103, so this will fail because the branch no longer exists. The `Create` method uses `-b` on the initial attempt (creates a new branch), but the retry path on line 43/50 uses bare `branchName` (checks out existing branch). After `cleanStaleBranch` deletes the branch, the retry on line 50 will fail with "branch not found". This needs a test to verify the full cleanup-and-retry path works end to end.

2. **Dispatch error doesn't reset work item status** (`work_item_handlers.go:247-258`). When `handleDispatch` fails at `s.sessions.Dispatch()`, the work item status has already been set to `InProgress` (line 247). The error is now properly surfaced to the UI, but the item stays stuck in `InProgress` with no running session. Should roll back the status on dispatch failure.

### Minor

3. **Toast timeout isn't cleaned up on unmount** (`Toast.tsx:29-31`). The `setTimeout` in `addToast` isn't tracked or cleared. If the provider unmounts before 6 seconds (unlikely in this app, but still), it'll attempt a state update on an unmounted component. Low risk given this wraps the entire app.

4. **`request_changes` button now shows `reviewError` twice** (`Panel.tsx:153-154`). The diff adds `{reviewError && <span ...>}` to the `request_changes` case, but `post_review` already shares the same `reviewError` state. If one of them sets the error, both button locations render it. Minor visual duplication but worth noting.

5. **Missing ARIA for toast dismiss button** (`Toast.tsx:49`). The close button uses `&times;` with no `aria-label`. Should have `aria-label="Dismiss"` for screen readers.

## Suggestions

- For finding #1: after `cleanStaleBranch`, the retry in `Create` should use `-b branchName` (recreate) rather than bare checkout, since the branch was just deleted. Or restructure so `cleanStaleBranch` doesn't delete the branch when the worktree removal succeeds (only delete when stale).
- For finding #2: add a rollback in the dispatch error path:
  ```go
  sess, err := s.sessions.Dispatch(r.Context(), item, skPtr, profile)
  if err != nil {
      s.store.UpdateWorkItemStatus(id, model.StatusPending)
      writeError(w, http.StatusInternalServerError, err.Error())
      return
  }
  ```

## Verdict

**Request changes.** The stale branch cleanup logic in `cleanStaleBranch` will delete the branch and then the caller tries to check out a branch that no longer exists -- this is the exact scenario this PR is meant to fix, so it needs to be right. The status rollback on dispatch failure is also important since the item gets stuck in a bad state. The frontend error surfacing is solid and well-structured.

jasonmadigan (commented):
Good. I've read through the full diff across all 12 files. Here's my review:

---

## CI Status

## Summary

Surfaces backend errors to the user in the UI through three mechanisms: (1) structured JSON error responses from all Go API handlers, (2) inline error display in modals (dispatch, new item), and (3) a new toast notification system for transient errors from session actions. Also adds stale worktree branch recovery in the backend so that "branch already exists" failures self-heal.

## Findings

**No critical issues.**

**Important:**

1. **`cleanStaleBranch` force-removes a worktree that may be in active use** (`worktree.go:94`). The method is called when a branch is "already checked out", but this can also happen when a *running* session holds the worktree. The comment says "stale" but there's no staleness check -- it just force-removes whatever worktree has that branch. If a second dispatch targets the same branch while the first session is still running, this will blow away the first session's working directory. Consider checking whether the worktree directory actually exists and is reachable before force-removing (truly stale = directory is gone or unreachable).

2. **Dispatch status rollback ignores its own error** (`work_item_handlers.go:258`). `s.store.UpdateWorkItemStatus(id, model.StatusReady)` on the rollback path doesn't check the returned error. If this fails, the item is stuck in "in_progress" with no session and no way to recover from the UI. Worth at least logging the error.

**Minor:**

3. **`handleDispatchSubmit` doesn't close the modal on success path timing** (`App.tsx:132`). If `api.dispatch` succeeds but `loadItems()` takes time, the modal closes and `loadItems` runs asynchronously. This is fine but `setShowDispatch(false)` and `loadItems()` could race with `DispatchModal`'s state -- since the component unmounts, the pending `setSubmitting(false)` in the modal's `handleSubmit` would fire on an unmounted component. React 18+ suppresses this warning but it's worth noting.

4. **Toast hardcoded border colours** (`styles.css:1337,1341`). `#991b1b` and `#1e3a5f` are hardcoded while the rest of the design uses CSS custom properties. Should match the existing pattern for maintainability.

5. **`writeError` is defined in `session_handlers.go` but used across all handler files** (via package-level visibility). Not a problem functionally, but it would read more naturally in a shared file like `server.go` or next to `writeJSON` if that were extracted.

## Suggestions

- For finding #1, add a guard in `cleanStaleBranch` that checks whether the worktree path actually exists on disk. If it does and looks healthy, don't force-remove it -- only clean up truly stale entries (directory gone, `.git` file broken).
- For finding #2, a one-liner: `if rollbackErr := s.store.UpdateWorkItemStatus(...); rollbackErr != nil { log.Printf("failed to roll back status: %v", rollbackErr) }`.
- For finding #4, extract those border colours to CSS custom properties like `--tint-red-border` / `--tint-blue-border`.

## Verdict

**Approve.** The core goal -- making errors visible to the user -- is well executed across both frontend and backend. The `writeError` migration is thorough, the toast system is clean, and the stale worktree recovery is a useful addition. Finding #1 is the most worth addressing (could cause data loss on concurrent dispatches to the same branch) but is an edge case that can be hardened in a follow-up.

PR #55, opened 2026-04-12
Drop-in should continue the agent's session, not start fresh

### LINKED ISSUE #12: Drop-in should continue the agent's session, not start fresh

jasonmadigan:
The 'copy cd + claude' button currently gives `cd <worktree> && claude` which starts a brand new session with no context from the agent's work.

It should use `claude --continue` or `claude --resume <session-id>` so the user picks up where the agent left off, with full conversation history and context.

The agent's session ID is available from our database. Two options:
- `cd <worktree> && claude --continue` -- continues the most recent session in that directory (simplest, should work if session was persisted)
- `cd <worktree> && claude --resume <session-id>` -- explicitly resumes by ID (more reliable)

Need to verify that sessions spawned with `--print` are persisted and resumable. If `--no-session-persistence` is not set, they should be.

Closes jasonmadigan/clawdio#12


jasonmadigan (commented):
Here's the review:

---

## CI Status

Passing (build-and-test: pass).

## Summary

Captures the Claude session ID from the `stream-json` init event and stores it in the database, then uses it in the frontend "copy cd + claude" button to produce `claude --resume <session-id>` instead of a bare `claude`, so users drop into the agent's existing conversation history.

## Findings

No critical or important issues. This is a clean, minimal change that threads a new field through all the right layers (model, store, migration, session manager, frontend types, UI).

**Minor:**

1. **`model.go:79` -- alignment:** `ClaudeSessionID` breaks the existing column alignment because the field name is longer. Go doesn't require aligned struct tags, but the rest of the struct is consistently aligned. Not a blocker, just slightly inconsistent visually.

2. **No test for `UpdateClaudeSessionID`:** The other store operations likely have tests; this new method doesn't appear to. Low risk given its simplicity (`UPDATE ... SET ... WHERE`), but worth noting.

## Observations

- The fallback to `--continue` when no session ID is captured is a sensible default. Covers edge cases where the init event might not include a session ID.
- Migration is additive (`ALTER TABLE ADD COLUMN`), consistent with the existing pattern.
- The session ID extraction from the init event aligns with the documented protocol (`session-protocol.md` line 22: "Session started, contains `session_id`").

## Verdict

**Approve.** Small, focused change that does exactly what the issue describes. No bugs, no edge case gaps, no style issues worth blocking on.

Issue #57, opened 2026-04-12
Session output should mimic Claude CLI terminal style


jasonmadigan:
The agent output area renders events as a flat stream of text and tool names. Claude's actual TTY output is structured: tool calls have headers, output is nested underneath, diffs are syntax-highlighted, and the visual hierarchy makes it easy to follow what the agent is doing.

## Current state

- Tool calls render as a single line with the tool name (e.g. `Read /path/to/file`)
- Text output is flat, unstyled `pre` text
- Tool input JSON is captured in `OutputEntry.detail` but never rendered
- No grouping -- text that follows a tool call isn't visually connected to it
- No diff rendering for Edit/Write operations
- Thinking text is italic but otherwise unstyled

## Desired behaviour

Match Claude CLI's terminal rendering:

**Tool call headers:**
- Prefix with a marker (Claude uses `⏺`)
- Show tool name and key input (file path, command, pattern)
- For Bash: show `$ command` with the actual command

**Tool output nesting:**
- Output that follows a tool call should be visually nested/indented under it
- Claude uses `⎿` as a connector character for continuation lines

**Content-aware rendering:**
- Bash output: monospace, dim
- Edit/Write: show the diff with +/- line colouring (green/red)
- Read: show file path, maybe line range
- Grep: show matches with file paths
- Errors: red, with `✘` marker
- Info/warnings: with `ℹ` marker

**Text (assistant messages):**
- Render markdown (already using `marked` elsewhere in the app)
- Clear visual separation from tool output

## What exists

- `OutputEntry` has `type` ('text' | 'tool' | 'thinking'), `content`, and `detail` (tool input JSON, currently unused)
- Server already extracts key info for common tools in `manager.go:457-475` (file paths for Read/Edit/Write/Glob/Grep, commands for Bash)
- Tool input JSON is sent via WebSocket but discarded by the renderer
- CSS is in `ui/src/styles.css` under `.agent-output`, `.output-tool`, `.output-text`, `.output-thinking`

## Implementation notes

- The server-side tool detail extraction (`manager.go:457-475`) could be extended to include the tool category (file_op, shell, search) so the frontend knows how to render each one
- Tool results (the output of a tool call) aren't currently sent as separate events -- they come back as `text` in the next assistant turn. May need a `tool_result` event type from the server, or group client-side by position
- Previous sessions show `s.summary` in a `<pre>` -- these could also benefit from basic formatting

PR #60, opened 2026-04-12
tmux session wrapping: peek, persistence, resume, remote operation

### LINKED ISSUE #58: tmux session wrapping: peek, persistence, resume, remote operation
Wrap dispatched agent sessions in tmux instead of bare subprocess exec. This solves several problems at once and sets up for headless/remote operation.

## What changes

Currently: `exec.Command("claude", "--output-format", "stream-json", ...)` with stdout piped directly.

Proposed: `tmux new-session -d -s clawdio-<short-id> claude --output-format stream-json ...` with output consumed via the JSONL session files or tmux pipe-pane.

## What this enables

**Peek (read-only live view):**
- `tmux attach -t clawdio-<id> -r` gives genuine read-only terminal access to a running agent
- UI gets a "Peek" button that copies the attach command
- See the full Claude CLI experience (diffs, tool calls, formatting) without interfering

**Session persistence:**
- Agent sessions survive clawdio process restarts -- tmux keeps them alive
- On startup, clawdio scans for existing `clawdio-*` tmux sessions and reconnects
- No more orphaned agent processes when the server crashes or restarts

**Resume/drop-in:**
- `tmux attach -t clawdio-<id>` is the natural drop-in (replaces the current `cd <worktree> && claude` approach from #12)
- You're literally in the same terminal, full context preserved
- Read-write attach for when you want to take over, read-only for when you just want to watch

**Remote/headless operation:**
- clawdio running on a remote machine (or in a container) with agents in tmux sessions
- SSH in, tmux attach to any running agent
- Container image: gh + tmux + claude + clawdio + API keys
- Multiple users could observe the same agent session simultaneously

## Output streaming

Two options for getting agent output into the UI:

1. **tmux pipe-pane** -- pipes terminal output to a file or command. We could pipe to a named pipe that the Go server reads. Gives us the rendered terminal output.
2. **JSONL tail** -- Claude Code writes session transcripts to `~/.claude/projects/<encoded-cwd>/<session-id>.jsonl`. Tail that file for structured events. This is what we currently get via stdout, just from a different source.
3. **Hybrid** -- use JSONL for structured events (what we stream to the UI now), tmux for the real terminal view (peek).

Option 3 is probably right. Keep the WebSocket streaming as-is (just change the source from stdout pipe to JSONL tail), and add tmux attach as the "full experience" peek.

## Implementation sketch

- `internal/session/manager.go`: replace `exec.Command` with tmux wrapping
- Session naming convention: `clawdio-<short-item-id>` (e.g. `clawdio-0b4aceed`)
- On startup: `tmux list-sessions -F '#{session_name}'` to find existing sessions
- Reconnect logic: match tmux session names back to work items, update DB status
- New API endpoint: `GET /api/sessions/{id}/peek` returns the tmux attach command
- UI: "Peek" button on active sessions (next to existing "copy cd" buttons)
- Cleanup: `tmux kill-session -t <name>` when a session completes and worktree is cleaned up

## Dependencies

tmux must be installed. For the target deployment (dev machines, remote servers, containers), this is not a real constraint. Could add a startup check that warns if tmux is not found and falls back to the current bare exec mode.

## Related

- #12 (drop-in should continue session) -- tmux attach replaces the `claude --resume` approach
- #57 (session output styling) -- tmux peek gives the real CLI experience; #57 improves the in-UI view
- #9 (bidirectional sessions) -- tmux attach is inherently bidirectional

jasonmadigan:
Wrap dispatched agent sessions in tmux instead of bare subprocess exec. This solves several problems at once and sets up for headless/remote operation.

## What changes

Currently: `exec.Command("claude", "--output-format", "stream-json", ...)` with stdout piped directly.

Proposed: `tmux new-session -d -s clawdio-<short-id> claude --output-format stream-json ...` with output consumed via the JSONL session files or tmux pipe-pane.

## What this enables

**Peek (read-only live view):**
- `tmux attach -t clawdio-<id> -r` gives genuine read-only terminal access to a running agent
- UI gets a "Peek" button that copies the attach command
- See the full Claude CLI experience (diffs, tool calls, formatting) without interfering

**Session persistence:**
- Agent sessions survive clawdio process restarts -- tmux keeps them alive
- On startup, clawdio scans for existing `clawdio-*` tmux sessions and reconnects
- No more orphaned agent processes when the server crashes or restarts

**Resume/drop-in:**
- `tmux attach -t clawdio-<id>` is the natural drop-in (replaces the current `cd <worktree> && claude` approach from #12)
- You're literally in the same terminal, full context preserved
- Read-write attach for when you want to take over, read-only for when you just want to watch

**Remote/headless operation:**
- clawdio running on a remote machine (or in a container) with agents in tmux sessions
- SSH in, tmux attach to any running agent
- Container image: gh + tmux + claude + clawdio + API keys
- Multiple users could observe the same agent session simultaneously

## Output streaming

Two options for getting agent output into the UI:

1. **tmux pipe-pane** -- pipes terminal output to a file or command. We could pipe to a named pipe that the Go server reads. Gives us the rendered terminal output.
2. **JSONL tail** -- Claude Code writes session transcripts to `~/.claude/projects/<encoded-cwd>/<session-id>.jsonl`. Tail that file for structured events. This is what we currently get via stdout, just from a different source.
3. **Hybrid** -- use JSONL for structured events (what we stream to the UI now), tmux for the real terminal view (peek).

Option 3 is probably right. Keep the WebSocket streaming as-is (just change the source from stdout pipe to JSONL tail), and add tmux attach as the "full experience" peek.

## Implementation sketch

- `internal/session/manager.go`: replace `exec.Command` with tmux wrapping
- Session naming convention: `clawdio-<short-item-id>` (e.g. `clawdio-0b4aceed`)
- On startup: `tmux list-sessions -F '#{session_name}'` to find existing sessions
- Reconnect logic: match tmux session names back to work items, update DB status
- New API endpoint: `GET /api/sessions/{id}/peek` returns the tmux attach command
- UI: "Peek" button on active sessions (next to existing "copy cd" buttons)
- Cleanup: `tmux kill-session -t <name>` when a session completes and worktree is cleaned up

## Dependencies

tmux must be installed. For the target deployment (dev machines, remote servers, containers), this is not a real constraint. Could add a startup check that warns if tmux is not found and falls back to the current bare exec mode.

## Related

- #12 (drop-in should continue session) -- tmux attach replaces the `claude --resume` approach
- #57 (session output styling) -- tmux peek gives the real CLI experience; #57 improves the in-UI view
- #9 (bidirectional sessions) -- tmux attach is inherently bidirectional

Closes jasonmadigan/clawdio#58


jasonmadigan (commented):
I've reviewed all the changes. Here's my review:

---

## CI Status

Passing.

## Summary

Wraps dispatched Claude sessions in tmux instead of bare subprocess exec, enabling peek (read-only attach), session persistence across server restarts, and drop-in via `tmux attach`. Falls back to bare exec when tmux is not installed. Adds a new `internal/tmux` package, a `/api/sessions/{id}/peek` endpoint, and UI buttons for peek/attach.

## Findings

### Critical

**1. Recovery logic is a no-op -- `CleanupOrphanedSessions` runs first and changes the status, so `RunningSessionsWithTmux` always returns empty.**

`main.go` calls `CleanupOrphanedSessions()` first (marks all running/starting sessions as `errored`), then calls `RecoverTmuxSessions()`. But `RunningSessionsWithTmux()` queries `WHERE status IN ('running', 'starting')` -- those rows no longer exist because cleanup just moved them to `errored`. Recovery will never find anything.

Fix: either query for `errored` sessions with a tmux name and a specific summary (e.g. `server restarted`), or run recovery *before* cleanup and have cleanup skip sessions that were successfully recovered.

**2. `resumeTmuxSession` always reports `SessionCompleted` -- never detects errors.**

`manager.go` lines ~348-352:
```go
status := model.SessionCompleted
exitCode := 0

if status == model.SessionCompleted {  // always true
```

The `if status == model.SessionCompleted` check is tautological -- `status` was just set to `SessionCompleted` on the previous line. There's no logic to detect whether the recovered tmux session actually errored. Compare with `runTmux` which has the same issue (though less severely, since it at least checks `ctx.Err()`).

### Important

**3. Shell injection in `PipePaneToFile`.**

`tmux.go:113`:
```go
cmd := exec.Command("tmux", "pipe-pane", "-t", name, "-o", fmt.Sprintf("cat >> %s", filePath))
```

The `filePath` is interpolated into a shell command string that tmux will execute. While the current callers use `os.TempDir()` + a controlled prefix, a path containing spaces or shell metacharacters would break or be exploitable. Use shell quoting (`'%s'` with single quotes) or avoid the shell entirely by writing a wrapper.

**4. `runTmux` uses `context.Background()` but `resumeTmuxSession` uses `m.ctx`.**

`runTmux` creates its context with `context.WithCancel(context.Background())` (line 398), meaning a server shutdown via `m.ctx` cancellation won't propagate to tmux sessions started this way. `resumeTmuxSession` correctly uses `m.ctx`. `runBare` also uses `context.Background()`. This looks like it predates the PR, but the tmux path makes it more visible -- a shutting-down server won't cleanly stop tailing tmux output.

**5. `tailOutput` polling with `time.Sleep(50ms)` and scanner recreation.**

Re-creating the `bufio.Scanner` on each poll iteration to pick up new data past EOF works, but is fragile. If the file is written to between the scanner hitting EOF and the `os.File` seek position advancing, data could be missed or duplicated. A `fsnotify` watcher or `tail -f` style approach (seeking to the last known position) would be more reliable. Not a blocker for merging, but worth a follow-up.

### Minor

**6. `copyDropIn` in `web/app.js` is now dead code.**

The old `copyDropIn('${esc(wtPath)}')` call in the active-session branch was replaced by `copyPeek`/`copyAttach`, but `copyDropIn` is still defined and used as a fallback in both `copyPeek` and `copyAttach`. This is fine for now, but the non-tmux fallback path (`copyDropIn`) references `item.repo_path` in `copyPeek` and does a second API call in `copyAttach` -- these fallback paths have different behaviour, which is a bit inconsistent.

**7. No error displayed to user when peek/attach API call fails.**

`copyPeek` and `copyAttach` silently fall back if the API returns no data. A failed fetch (network error, 500) will also silently do nothing.

## Suggestions

1. Fix the recovery ordering -- simplest approach: change `RunningSessionsWithTmux` to query `WHERE status = 'errored' AND tmux_session_name != '' AND summary = 'server restarted'` since that's the state cleanup leaves them in.
2. Add actual error detection in `resumeTmuxSession` -- at minimum check if the tmux session died while tailing (it does, via `SessionExists`, but doesn't set status to errored).
3. Quote the file path in `PipePaneToFile`: `fmt.Sprintf("cat >> '%s'", filePath)` or better, use `exec.Command("tee", "-a", filePath)` piped differently.
4. Consider using `m.ctx` as the parent context in `runTmux` for clean shutdown.

## Verdict

**Request changes.** The tmux package, session manager refactoring, and UI changes are well-structured and the test coverage is solid. But the recovery logic (#1) is the headline feature of this PR and it's currently broken -- sessions will never be recovered after a restart. The tautological status check (#2) means recovered sessions would always be marked completed even if they errored. These need fixing before merge.

jasonmadigan (commented):
Here's the review:

---

## CI Status

No CI checks configured on this branch. Can't verify build/tests pass automatically.

## Summary

Wraps dispatched agent sessions in tmux instead of bare subprocess exec. Adds peek (read-only attach), session persistence across server restarts, and drop-in via tmux attach. Falls back to bare exec when tmux isn't installed. Clean separation into a new `internal/tmux` package with good test coverage.

Well-structured PR. The second commit addresses earlier review feedback (recovery ordering, error detection, shell quoting, context propagation), which shows good iteration. Two issues remain.

## Findings

### Important

**1. Missing crash detection in `runTmux` -- `internal/session/manager.go:457-465`**

`resumeTmuxSession` (lines 352-360) correctly detects when a tmux session vanished without producing output:

```go
} else if !tmux.SessionExists(sess.TmuxSessionName) && lastText == "" {
    status = model.SessionErrored
    exitCode = 1
}
```

But `runTmux` (lines 457-465) lacks this check. If the claude process crashes inside tmux without emitting a `"result"` event, `tailOutput` returns normally (the tmux session is gone, so the polling loop exits), and the session gets marked `SessionCompleted` with exit code 0. This would trigger `AutoCommit` on a potentially broken worktree.

Fix: add the same guard after line 465:

```go
if ctx.Err() != nil {
    tmux.KillSession(tmuxName)
    status = model.SessionErrored
    exitCode = -1
} else if !tmux.SessionExists(tmuxName) && lastText == "" {
    status = model.SessionErrored
    exitCode = 1
}
```

### Minor

**2. Single-quote shell escaping in `PipePaneToFile` -- `internal/tmux/tmux.go:110`**

`cat >> '%s'` uses single-quote escaping, which breaks if the path itself contains a single quote. The path is constructed from `os.TempDir()` + session UUID prefix, so this is safe in practice, but it's worth noting. If you ever change the path construction, this becomes a real injection vector. A comment noting the constraint would suffice, or use double-quoting with proper escaping.

## What looks good

- Recovery flow is correct: `CleanupOrphanedSessions` marks sessions as errored with `summary='server restarted'`, then `RecoverableSessionsWithTmux` queries exactly that, cross-referenced against live tmux sessions. Well thought through.
- The `processLine` extraction is a clean refactor -- same logic shared between bare and tmux modes.
- tmux package is well-bounded: pure wrapper, no business logic, properly handles "no server running" edge cases.
- Tests cover the store layer and tmux operations (with `t.Skip` for environments without tmux).
- Graceful fallback when tmux is unavailable.

## Verdict

**Request changes** -- the crash detection gap in `runTmux` is a real bug that would silently mark crashed sessions as completed and auto-commit potentially broken state. One-line fix.

jasonmadigan (commented):
CI is passing. I've read the full diff. Here's my review:

---

## CI Status

Passing (build-and-test, 33s).

## Summary

Wraps dispatched agent sessions in tmux instead of bare subprocess exec. Adds a new `internal/tmux` package, pipe-pane-based output tailing, session recovery on server restart, a `/api/sessions/{id}/peek` endpoint, and peek/attach buttons in the UI. Falls back to bare exec when tmux is unavailable.

## Findings

### Important

1. **`SendKeys` for stdin is problematic** (`internal/tmux/tmux.go:57-63`, `internal/session/manager.go:154-156`). `tmux send-keys -l` sends text as keyboard input to the terminal -- it doesn't write to the process's stdin pipe. This means:
   - Special characters in JSON payloads (braces, quotes, colons) could interact badly with terminal state depending on what's running
   - There's a race between keystrokes being processed and the process reading them
   - For `control_request` approve/deny messages, this is the critical path. If Claude is mid-output when keys arrive, the JSON message could get interleaved with terminal escape sequences
   
   Consider using `tmux send-keys` with `Enter` appended, or better, writing to a named pipe/fifo that the process reads from. At minimum, this needs testing with real approve/deny flows.

2. **`PipePaneToFile` shell injection surface** (`internal/tmux/tmux.go:117-122`). The `filePath` is interpolated into a shell command string: `fmt.Sprintf("cat >> '%s'", filePath)`. The comment notes the caller must ensure no single quotes in the path, but this is a fragile contract. If `os.TempDir()` ever returns something unexpected (or a future caller passes user input), this breaks. Consider using a format that doesn't need quoting, e.g. just passing the raw path since `os.TempDir()` paths are safe, or validating the path.

3. **Crash detection gap in `tailOutput`** (`internal/session/manager.go:441-499`). The polling loop checks `tmux.SessionExists` to detect when the session exits, but there's a window: if the tmux session crashes between the `SessionExists` check and the next `scanner.Scan()`, the loop could spin on EOF + sleep(50ms) indefinitely until the *next* `SessionExists` check catches it. This is minor (50ms window, low probability) but worth noting.

4. **`runBare` uses `context.Background()` while `runTmux` uses `m.ctx`** (`internal/session/manager.go:503` vs `:368`). The bare path creates its own context from `context.Background()`, meaning `CancelAll()` on server shutdown won't cancel bare sessions through the parent context. This is a pre-existing issue, but the inconsistency is now visible. Worth a follow-up to align them.

### Minor

5. **Duplicated completion logic** (`internal/session/manager.go`). The auto-commit + finish sequence appears in three places: `runTmux`, `resumeTmuxSession`, and `runBare`. Consider extracting a `completeSession(sessionID, worktreePath, lastText, ctxErr, tmuxName)` helper.

6. **`RecoverableSessionsWithTmux` relies on `summary = 'server restarted'`** (`internal/store/store.go:283`). This couples recovery to the exact string set by `CleanupOrphanedSessions`. If that string changes, recovery silently breaks. A constant would make this coupling explicit.

7. **`tailOutput` re-creates the scanner on each poll cycle** (`internal/session/manager.go:496-498`). This works because the file handle's position is preserved, but it allocates a new 1MB buffer each cycle. For long-running sessions, this is a lot of allocations. Could keep a single `bufio.Reader` and seek/read manually.

8. **No test for `processLine`**. The refactoring to extract `processLine` from inline scanning is good, but it's now a shared function used by both tmux and bare paths without its own unit test.

## Suggestions

- For finding #1: short-term, add a comment flagging `SendKeys` as a known limitation for bidirectional comms. Long-term, consider writing to a fifo or using `tmux load-buffer`/`paste-buffer` for atomic message delivery.
- For finding #5: extract a `finaliseSession` method to DRY up the three completion sites.
- For finding #6: define `const orphanedSummary = "server restarted"` in the store package and reference it from both `CleanupOrphanedSessions` and `RecoverableSessionsWithTmux`.

## Verdict

**Approve.** This is a solid feature that cleanly layers tmux wrapping on top of the existing exec path with a proper fallback. The architecture (pipe-pane for output, recovery on restart, peek endpoint) is well thought out. The `SendKeys` concern (#1) is the main thing to watch -- it'll work for simple cases but may need revisiting when bidirectional comms become more critical. The duplication (#5) and magic string (#6) are easy follow-ups.

jasonmadigan (commented):
I've reviewed the full diff. Here's my review:

---

## CI Status

No CI checks configured for this branch.

## Summary

Wraps dispatched agent sessions in tmux instead of bare `exec.Command`, enabling peek (read-only tmux attach), session persistence across server restarts, and a clean drop-in path. Falls back to direct subprocess when tmux isn't installed. Adds a new `internal/tmux` package, a `/api/sessions/{id}/peek` endpoint, DB migration for `tmux_session_name`, recovery logic on startup, and UI buttons for peek/attach.

## Findings

### Important

1. **`runBare` uses `context.Background()`, `runTmux` uses `m.ctx` -- inconsistent context parenting** (`manager.go:413` vs `manager.go:528`). `runBare` creates a context from `context.Background()`, so `CancelAll()` (which calls `rs.cancel()`) still works, but the manager's own shutdown context (`m.ctx`) won't propagate to bare sessions. This is the pre-existing behaviour, but now that `runTmux` correctly parents from `m.ctx`, the asymmetry is worth noting. If the manager's context is cancelled (e.g. SIGTERM), bare sessions won't get cancelled through that path. Consider parenting both from `m.ctx` for consistency.

2. **`SendKeys` for JSON input is fragile** (`tmux.go:63-70`). The doc comment acknowledges this, which is good. `send-keys -l` writes to the terminal emulator, not the process stdin. If the claude process is mid-output, terminal echo or line-editing state could corrupt the message. This is fine as a first pass since control requests (approve/deny) are infrequent, but worth tracking as a known limitation for bidirectional use (#9).

3. **`tailOutput` re-creates `bufio.Scanner` in a polling loop** (`manager.go:520-521`). When the scanner hits EOF and a new scanner is created on the same `*os.File`, it works because the file offset is maintained on the `*os.File`. But each new scanner allocates a fresh 1MB buffer. At 50ms polling interval that's a lot of allocations. A lighter approach: use a single `bufio.Reader` and `ReadLine`/`ReadBytes('\n')` in the loop, which naturally blocks at EOF and retries on the same buffer.

4. **`pipe-pane` captures rendered terminal output, not raw stdout** (`tmux.go:116-117`). `pipe-pane` gets the output _after_ the terminal emulator processes it, which means ANSI escape sequences, cursor movements, and terminal control codes end up in the file. Claude's `--output-format stream-json` should produce clean JSONL on stdout, but if tmux's PTY processing injects anything (line wrapping, etc.), the JSONL parser will see unparseable lines. The `processLine` function handles parse errors gracefully (logs and continues), so this won't crash, but could cause dropped events. Worth monitoring in practice.

### Minor

5. **Output file path collision on recovered sessions** (`manager.go:316`). `resumeTmuxSession` constructs the output file path using `sess.ID[:8]`, same as `runTmux`. If a session is recovered and somehow the old output file from a previous run still exists in `/tmp`, `pipe-pane` appends to it (`cat >> ...`), which means the recovered tail could re-process stale lines from the previous run. The `processLine` function is stateful (dedup via `seenToolIDs`, cumulative text lengths), so duplicated init events or assistant messages from the old run could produce spurious UI events. Consider truncating the file before setting up pipe-pane on recovery, or using a unique suffix.

6. **`PipePaneToFile` shell injection surface** (`tmux.go:115-117`). The comment notes single-quote safety, but the mitigation relies on callers never passing a path with single quotes. The actual callers use `os.TempDir()` + a fixed prefix, so this is safe in practice. Just flagging for awareness if the function is ever reused with user-supplied paths.

7. **No test for `processLine` or `tailOutput` in isolation**. The refactoring to extract `processLine` from the inline scanner loop is good, but there's no unit test covering the extracted function. The tmux integration tests cover the plumbing, and the store tests cover recovery queries, but the event parsing/dispatch logic is only tested implicitly via integration.

## Suggestions

- For (1): one-line fix, change `runBare` to `context.WithCancel(m.ctx)` for parity.
- For (3): replace the scanner-recreation loop with a `bufio.Reader` that persists across EOF. Reduces allocations significantly.
- For (5): add `os.Remove(outputFile)` before `tmux.PipePaneToFile` in `resumeTmuxSession` to ensure a clean file.

## Verdict

**Approve.** The design is sound -- tmux wrapping with bare fallback is the right architecture, the recovery logic is well thought through, and the extraction of `processLine` makes the code cleaner. The issues above are all minor-to-moderate and none are blocking. The `send-keys` limitation and `pipe-pane` output fidelity are acknowledged in the code comments, which is the right call for a first pass. Tests are solid -- both unit tests for the tmux package and integration-style tests for the store recovery path.

PR #61, opened 2026-04-13
Session output should mimic Claude CLI terminal style

### LINKED ISSUE #57: Session output should mimic Claude CLI terminal style

jasonmadigan:
The agent output area renders events as a flat stream of text and tool names. Claude's actual TTY output is structured: tool calls have headers, output is nested underneath, diffs are syntax-highlighted, and the visual hierarchy makes it easy to follow what the agent is doing.

## Current state

- Tool calls render as a single line with the tool name (e.g. `Read /path/to/file`)
- Text output is flat, unstyled `pre` text
- Tool input JSON is captured in `OutputEntry.detail` but never rendered
- No grouping -- text that follows a tool call isn't visually connected to it
- No diff rendering for Edit/Write operations
- Thinking text is italic but otherwise unstyled

## Desired behaviour

Match Claude CLI's terminal rendering:

**Tool call headers:**
- Prefix with a marker (Claude uses `⏺`)
- Show tool name and key input (file path, command, pattern)
- For Bash: show `$ command` with the actual command

**Tool output nesting:**
- Output that follows a tool call should be visually nested/indented under it
- Claude uses `⎿` as a connector character for continuation lines

**Content-aware rendering:**
- Bash output: monospace, dim
- Edit/Write: show the diff with +/- line colouring (green/red)
- Read: show file path, maybe line range
- Grep: show matches with file paths
- Errors: red, with `✘` marker
- Info/warnings: with `ℹ` marker

**Text (assistant messages):**
- Render markdown (already using `marked` elsewhere in the app)
- Clear visual separation from tool output

## What exists

- `OutputEntry` has `type` ('text' | 'tool' | 'thinking'), `content`, and `detail` (tool input JSON, currently unused)
- Server already extracts key info for common tools in `manager.go:457-475` (file paths for Read/Edit/Write/Glob/Grep, commands for Bash)
- Tool input JSON is sent via WebSocket but discarded by the renderer
- CSS is in `ui/src/styles.css` under `.agent-output`, `.output-tool`, `.output-text`, `.output-thinking`

## Implementation notes

- The server-side tool detail extraction (`manager.go:457-475`) could be extended to include the tool category (file_op, shell, search) so the frontend knows how to render each one
- Tool results (the output of a tool call) aren't currently sent as separate events -- they come back as `text` in the next assistant turn. May need a `tool_result` event type from the server, or group client-side by position
- Previous sessions show `s.summary` in a `<pre>` -- these could also benefit from basic formatting

Closes jasonmadigan/clawdio#57


jasonmadigan (commented):
Here's the review:

---

## CI Status

Passing (build-and-test).

## Summary

Refactors the agent output renderer from flat text into a structured, CLI-style view. Tool calls get category-aware headers with markers, text following a tool call is visually nested underneath, Edit/Write operations show diffs with green/red colouring, and assistant text is rendered as markdown. The Go backend is extended to classify tools by category (`file_op`, `shell`, `search`, `agent`, `other`) and pass the raw tool name through to the frontend.

7 files changed, +335 / -46.

## Findings

### Important

**XSS via `dangerouslySetInnerHTML` on unsanitised `marked` output** (`AgentOutput.tsx:92-99`)

`marked` v18 does not sanitise HTML by default. `entry.content` comes from Claude's assistant text responses, which can contain raw HTML. While the threat model here is limited (the content source is Claude, not arbitrary user input), passing `marked.parse()` straight into `dangerouslySetInnerHTML` without sanitisation is a risk if the output ever includes injected content (e.g. from tool results that get echoed as assistant text). The existing uses in `Panel.tsx` and `SkillsView.tsx` have the same issue, so this is pre-existing rather than newly introduced -- but worth noting since you're adding another instance.

Consider adding DOMPurify or using `marked`'s extension API to strip raw HTML tags.

**Diff rendering only shows old/new blocks, not an actual diff** (`AgentOutput.tsx:103-122`)

`renderDiff` dumps all removed lines then all added lines. For edits that change a few lines in a larger block, this loses the positional context that makes diffs useful. It also won't handle `Write` well -- a full `new_string` with no `old_string` would render the entire file content as additions. For a first pass this is fine, but it's worth knowing the limitation.

### Minor

**`json.Unmarshal` error silently ignored** (`manager.go:518`)

`toolDetail` ignores the error from `json.Unmarshal`. If `inputJSON` is malformed, `inp` will be nil and the function falls through to the default case returning just the name -- which is reasonable degradation. But a debug log on unmarshal failure would help when troubleshooting why tool details aren't rendering.

**Bash command truncation changed from 80 to 120 chars** (`manager.go:540`)

The old code truncated at 80 chars. Now it's 120. Not a problem, just noting the behavioural change. Long commands in a narrow panel might wrap awkwardly.

**`groupEntries` creates a new group per tool call** (`AgentOutput.tsx:15-33`)

If Claude emits two tool calls in a row with no text between them (common with parallel tool calls), each gets its own group with an empty `children` array. This works correctly but means there's no visual indication that those tool calls were part of the same batch. Not a bug, just a potential future improvement.

**`Glob` categorised as `file_op` but uses `pattern` not `file_path`** (`manager.go:521-528`)

The switch groups `Glob` with `Read`/`Edit`/`Write` under `file_op`. The fallback to `inp["pattern"]` handles it, but Glob is arguably closer to `search` than `file_op` -- it searches for files by pattern rather than operating on a specific file. Minor semantic quibble.

## Suggestions

- Consider DOMPurify for the `marked` output, or at minimum configure `marked` to strip raw HTML with a custom renderer. This applies to the existing uses too.
- The diff rendering would benefit from context lines (showing unchanged lines around changes) in a future iteration. For now, adding a "full file" indicator when `old_string` is absent would help distinguish writes from edits.

## Verdict

Approve. The refactoring is clean, the grouping logic is sound, and the new `toolDetail` function is a good extraction. The XSS concern is pre-existing rather than newly introduced, and the diff rendering limitation is a known tradeoff for a first pass. No blockers.

PR #70, opened 2026-04-13
Persist agent output by reading Claude Code session JSONL files

### LINKED ISSUE #67: Persist agent output by reading Claude Code session JSONL files
Agent output is currently held in React state only -- a page refresh wipes it. Instead of storing output ourselves, read it from Claude Code's own session transcripts.

## How Claude Code stores sessions

Claude writes JSONL transcripts to:
```
~/.claude/projects/<encoded-cwd>/<session-id>.jsonl
```

The path encoding replaces `/` with `-` and strips dots from directory names. For example:
```
worktree: /Users/jmadigan/.clawdio/worktrees/clawdio/72062ceb-...
encodes to: -Users-jmadigan--clawdio-worktrees-clawdio-72062ceb-...
```

We already store `claude_session_id` and `worktree_path` on each session, so we can derive the JSONL file path.

## Implementation

**Backend:**
- New function: derive Claude project path from worktree path (encode path separators, strip dots)
- New endpoint: `GET /api/sessions/{id}/output` -- reads the JSONL file, parses it using the same logic as `processLine`/`handleAssistantMessage`, and returns structured output entries
- The parsing logic already exists in `manager.go` -- extract it into a reusable function that can process a file on disk
- Return the same `OutputEntry` shape the frontend already understands (type, content, detail, rawName, category)

**Frontend:**
- On panel mount (or when selecting a work item with sessions), fetch `GET /api/sessions/{id}/output` for the latest session
- Use the returned entries as the initial `liveOutput` state
- WebSocket events append on top with a sequence watermark to avoid dupes
- Previous sessions section could also show structured output instead of just the summary text

## Edge cases

- JSONL file may not exist (session ran before claude_session_id tracking, or worktree was cleaned up) -- fall back to showing the summary text as today
- File may be large for long sessions -- consider pagination or streaming, or just cap at last N entries
- Worktree may have been deleted but the JSONL file persists in `~/.claude/projects/` -- this is fine, the file outlives the worktree
- Running sessions: the file is being actively written to. The endpoint should read what's available (same as the tmux tail approach)

## What this replaces

Currently `session.summary` stores the final assistant text as a flat string. With this change, we can show the full structured output for any session -- past or present -- with tool calls, diffs, thinking, the lot. The summary field remains useful as a quick preview on the board cards.

jasonmadigan:
Agent output is currently held in React state only -- a page refresh wipes it. Instead of storing output ourselves, read it from Claude Code's own session transcripts.

## How Claude Code stores sessions

Claude writes JSONL transcripts to:
```
~/.claude/projects/<encoded-cwd>/<session-id>.jsonl
```

The path encoding replaces `/` with `-` and strips dots from directory names. For example:
```
worktree: /Users/jmadigan/.clawdio/worktrees/clawdio/72062ceb-...
encodes to: -Users-jmadigan--clawdio-worktrees-clawdio-72062ceb-...
```

We already store `claude_session_id` and `worktree_path` on each session, so we can derive the JSONL file path.

## Implementation

**Backend:**
- New function: derive Claude project path from worktree path (encode path separators, strip dots)
- New endpoint: `GET /api/sessions/{id}/output` -- reads the JSONL file, parses it using the same logic as `processLine`/`handleAssistantMessage`, and returns structured output entries
- The parsing logic already exists in `manager.go` -- extract it into a reusable function that can process a file on disk
- Return the same `OutputEntry` shape the frontend already understands (type, content, detail, rawName, category)

**Frontend:**
- On panel mount (or when selecting a work item with sessions), fetch `GET /api/sessions/{id}/output` for the latest session
- Use the returned entries as the initial `liveOutput` state
- WebSocket events append on top with a sequence watermark to avoid dupes
- Previous sessions section could also show structured output instead of just the summary text

## Edge cases

- JSONL file may not exist (session ran before claude_session_id tracking, or worktree was cleaned up) -- fall back to showing the summary text as today
- File may be large for long sessions -- consider pagination or streaming, or just cap at last N entries
- Worktree may have been deleted but the JSONL file persists in `~/.claude/projects/` -- this is fine, the file outlives the worktree
- Running sessions: the file is being actively written to. The endpoint should read what's available (same as the tmux tail approach)

## What this replaces

Currently `session.summary` stores the final assistant text as a flat string. With this change, we can show the full structured output for any session -- past or present -- with tool calls, diffs, thinking, the lot. The summary field remains useful as a quick preview on the board cards.

Closes jasonmadigan/clawdio#67


jasonmadigan (commented):
Good. I've now read the full diff and surrounding context. Here's my review:

---

## CI Status

Passing (build-and-test).

## Summary

Reads Claude Code's session JSONL transcript files from disk to persist and restore structured agent output (text, tool calls, thinking blocks). Previously, output was held only in React state and lost on page refresh. The change extracts `toolDetail` into a new `claudefile` package, adds a `GET /api/sessions/{id}/output` endpoint, and loads persisted output in the frontend on session selection.

## Findings

### Important

1. **`WorktreePath` on sessions may differ from the worktree that was active when Claude wrote the JSONL** -- `claudefile.go:18-21`. The path encoding function (`ProjectDir`) assumes the worktree path stored on the session is exactly what Claude used to derive its project directory. If there are symlinks or path normalisation differences (e.g. `/var` vs `/private/var` on macOS), the derived JSONL path won't match. This is likely fine in practice since both clawdio and Claude Code see the same canonical path, but worth being aware of.

2. **`os.UserHomeDir` error is silently ignored** -- `claudefile.go:19`. `SessionJSONLPath` ignores the error from `os.UserHomeDir()`. If it fails (unlikely but possible in restricted environments), `home` is empty and the path is nonsensical. Consider returning an error or at least logging.

3. **Missing `liveOutputs` in the `useEffect` dependency array** -- `App.tsx:73`. The effect reads `liveOutputs[target.id]` but the lint suppression comment hides this. The current `if (prev[target.id] && prev[target.id].length > 0) return prev` guard inside `setLiveOutputs` prevents issues, but this means the persisted output fetch fires on every `items` change even when output is already loaded. The guard makes it a no-op, so it's functionally correct but creates unnecessary network requests.

4. **Large JSONL files will block the HTTP handler** -- `claudefile.go:33-60`. `ParseJSONL` reads the entire file synchronously. Long-running sessions can produce multi-megabyte JSONL files. The `maxEntries` param helps cap the response size but doesn't help with read/parse time since it still reads the whole file and then truncates. The PR description mentions this as a known edge case, so flagging it for awareness rather than as a blocker.

### Minor

5. **`OutputEntry` re-export in App.tsx looks unnecessary** -- `App.tsx:3`. `export type { OutputEntry }` re-exports the type from `types.ts`, but nothing in the diff imports `OutputEntry` from `App.tsx`. If other components already import from `types.ts`, this re-export is dead code. If existing consumers import from `App.tsx`, this is fine for backwards compat.

6. **`GetSession` error doesn't distinguish not-found from other DB errors** -- `session_handlers.go:98-101`. If `GetSession` fails for a reason other than "not found" (DB corruption, connection issue), the handler returns 404 which is misleading. The existing `handleSessionPeek` has the same pattern so this is consistent, but worth noting.

7. **Test file path in test uses hardcoded username** -- `claudefile_test.go:14`. The test case references `/Users/jmadigan/` which is fine for a unit test (it's just string manipulation), but ties the expected output to a specific user's path format.

## Suggestions

- For issue 2: change to `home, err := os.UserHomeDir(); if err != nil { return "" }` or return an error from `SessionJSONLPath`.
- For issue 3: consider adding a `fetchedOutputRef` to track which session IDs have already been fetched, avoiding repeat requests without needing `liveOutputs` in the dependency array.
- For the dead re-export (5): check if anything imports `OutputEntry` from `App.tsx`. If not, drop the re-export line.

## Verdict

**Approve.** Clean extraction of `toolDetail` into a shared package, solid test coverage for the JSONL parsing, and the endpoint + frontend integration is straightforward. The issues noted are minor/awareness-level -- none are blockers.

PR #68, opened 2026-04-13
Skill chaining: composable workflows from sequences of skills

jasonmadigan:
The orchestrator should understand multi-step workflows composed of skills. A workflow chains skills with transitions between them.

**Key design decision:** GitHub is the shared state between steps. Each skill reads from and writes to GitHub (comments, reviews, status). No local context passing needed between dispatch steps.

**Examples:**

```yaml
# implement-and-land
steps:
  - skill: implement-issue
    then: push_and_pr
  - skill: review-pr
    on_approve: merge
    on_changes_requested: address-feedback
  - skill: address-feedback
    then: push
    then: goto review-pr
```

**How state flows between steps:**
- implement-issue commits code, orchestrator pushes + creates PR
- review-pr reads the PR diff, posts review comments to GitHub
- address-feedback reads PR comments from GitHub, addresses them, pushes
- Each step reads GitHub to understand what happened before

**Implementation:**
- Workflows are declarative files (markdown/yaml) in `skills/`
- Each step is a regular skill dispatch (same worktree, same session model)
- Decision points: auto (always continue), approval (user clicks), conditional (based on output)
- Autopilot mode (#23) could auto-run entire workflows
- UI shows workflow progress: which step you're on, what's next

Closes jasonmadigan/clawdio#29

PR #71, opened 2026-04-13
Skill chaining: composable workflows from sequences of skills

### LINKED ISSUE #29: Skill chaining: composable workflows from sequences of skills
The orchestrator should understand multi-step workflows composed of skills. A workflow chains skills with transitions between them.

**Key design decision:** GitHub is the shared state between steps. Each skill reads from and writes to GitHub (comments, reviews, status). No local context passing needed between dispatch steps.

**Examples:**

```yaml
# implement-and-land
steps:
  - skill: implement-issue
    then: push_and_pr
  - skill: review-pr
    on_approve: merge
    on_changes_requested: address-feedback
  - skill: address-feedback
    then: push
    then: goto review-pr
```

**How state flows between steps:**
- implement-issue commits code, orchestrator pushes + creates PR
- review-pr reads the PR diff, posts review comments to GitHub
- address-feedback reads PR comments from GitHub, addresses them, pushes
- Each step reads GitHub to understand what happened before

**Implementation:**
- Workflows are declarative files (markdown/yaml) in `skills/`
- Each step is a regular skill dispatch (same worktree, same session model)
- Decision points: auto (always continue), approval (user clicks), conditional (based on output)
- Autopilot mode (#23) could auto-run entire workflows
- UI shows workflow progress: which step you're on, what's next

jasonmadigan:
Re-opened from #68 (branch was accidentally force-pushed to main ref).

Closes #29


jasonmadigan (commented):
Confirmed. The migration at index 6 (`ALTER TABLE workflow_runs ADD COLUMN total_steps`) will fail on existing databases where `workflow_runs` doesn't exist yet, because the error filter only handles "duplicate column" errors, not "no such table". The `CREATE TABLE IF NOT EXISTS workflow_runs` in the schema only runs for the initial creation; existing databases rely on migrations.

Now I have everything I need. Here's the review:

---

## CI Status

Passing (build-and-test: 32s).

## Summary

Adds a workflow engine that chains skills into multi-step sequences. A workflow YAML file declares steps (skill dispatches), post-actions (push, push_and_pr, merge), and transitions (on_approve, on_changes_requested, goto). The engine advances through steps automatically as sessions complete.

New packages: `internal/workflow` (parser, manager, engine), new API endpoints, new `workflow_runs` DB table, and UI components for workflow progress and dispatch.

## Findings

### Critical

1. **Migration will crash existing databases** (`internal/store/migrations.go:11`)

   Migration index 6 is `ALTER TABLE workflow_runs ADD COLUMN total_steps ...` but `workflow_runs` doesn't exist on databases created before this PR. The schema's `CREATE TABLE IF NOT EXISTS` only runs at initial DB creation. The migration error handler only ignores "duplicate column" errors, so "no such table" will cause a fatal startup error.

   Fix: add a `CREATE TABLE IF NOT EXISTS workflow_runs (...)` migration *before* the `ALTER TABLE` migration. Or collapse both into a single `CREATE TABLE IF NOT EXISTS` migration that includes `total_steps` from the start.

### Important

2. **`advanceWorkflow` runs in a goroutine with no error surface** (`internal/session/manager.go:868`)

   When `advanceWorkflow` fails (e.g. post-action fails, engine returns error), it silently logs and returns. The workflow run stays in `running` status indefinitely -- the UI will show it as active with no way to retry or diagnose. The `advanceWorkflow` goroutine should mark the run as `errored` on any terminal failure path, not just some of them.

   For example, the `ExecutePostAction` failure at line ~407 logs but doesn't update the run status to `errored`, leaving it stuck.

3. **Duplicated post-action + advancement logic** (`internal/session/manager.go:883-960` and `963-1060`)

   `advanceWorkflow` and `WorkflowReviewOutcome` both contain near-identical sequences: execute post-action, check for `pr_created:` prefix, call `PostActionDone`, then dispatch next skill. This is fragile -- a fix to one needs to be replicated in the other. Worth extracting a shared method.

4. **`WorkflowReviewOutcome` searches for worktree path by iterating all sessions** (`internal/session/manager.go:1014-1024`)

   This grabs the first session with a non-empty `WorktreePath`, which may not be the right worktree for this workflow step (e.g. if the item has had multiple dispatch sessions). Should filter by `WorkflowRunID` or at least pick the most recent session.

5. **Custom YAML parser silently ignores unknown structures** (`internal/workflow/parse.go`)

   Lines with no `:` separator are silently dropped (line 70: `continue`). A typo like `sskill: implement-issue` produces a step with an empty `Skill` field. The validation at line 122-125 catches missing skills, but the root cause (the typo) is invisible. Consider warning on unrecognised keys within a step.

### Minor

6. **`StepResult.Paused` is declared but never set** (`internal/workflow/engine.go:27`)

   The struct has a `Paused bool` field, but no code path ever sets it to `true`. Either remove it or wire it up.

7. **`WorkflowRun` type is missing `total_steps` in the TS interface** (`ui/src/lib/types.ts:96-103`)

   The Go model has `TotalSteps` and the DB stores it, but the TS `WorkflowRun` interface omits it. Not functionally broken since the UI doesn't display it, but the API response and type don't match.

8. **`WorkflowDropdown` renders N buttons in a flat span** (`ui/src/components/Panel.tsx:235-249`)

   If there are many workflows, this will overflow. A `<select>` or an actual dropdown menu would scale better. Fine for now with 1-2 workflows.

## Suggestions

- For finding #1: the simplest fix is to make migration index 6 a CREATE TABLE rather than ALTER TABLE. Since both this migration and the schema define the same table, the migration can be the canonical `CREATE TABLE IF NOT EXISTS workflow_runs (... including total_steps ...)` and the ALTER TABLE can be removed entirely.
- For finding #3: extract the shared post-action execution + advancement sequence into a helper like `executeAndAdvance(ctx, run, result, item, wtPath)`.
- For finding #4: filter the session query by `workflow_run_id = run.ID` rather than picking the first session with a worktree path.

## Verdict

**Request changes.** The migration issue (#1) will break any existing installation on upgrade. The other findings are important but not blocking.

jasonmadigan (commented):
I have everything I need. Here's my review:

---

## CI Status

Passing (build-and-test: 33s).

## Summary

Adds a workflow engine that chains skills into multi-step sequences. A YAML file declares steps with skill dispatches, post-actions (`push`, `push_and_pr`, `merge`), and conditional transitions (`on_approve`, `on_changes_requested`, `goto`). New packages: `internal/workflow` (parser, engine, manager), new API endpoints, `workflow_runs` DB table, and UI components for progress and dispatch.

## Previous Review

An earlier review (by the PR author) flagged 8 issues. Three follow-up commits (`0d02429`, `70f9eea`, `6ed56b8`) address them:

| Finding | Status |
|-|-|
| Migration crashes existing DBs | **Fixed.** Now `CREATE TABLE IF NOT EXISTS` instead of `ALTER TABLE` |
| `advanceWorkflow` no error surface | **Fixed.** `markErrored()` helper covers all failure paths |
| Duplicated post-action logic | **Fixed.** Extracted to `executeAndAdvance()` shared method |
| Worktree path picking wrong session | **Fixed.** Iterates in reverse, filters by `WorkflowRunID == runID` |
| Parser ignores unknown keys silently | **Fixed.** `log.Printf` for unrecognised keys in `parseStepKey()` |
| `StepResult.Paused` never set | **Fixed.** Field removed |
| TS `WorkflowRun` missing `total_steps` | **Fixed.** Added to the interface |
| `WorkflowDropdown` overflow | Acknowledged as acceptable for now |

All flagged issues have been addressed in the follow-up commits.

## Findings

### Important

1. **`TotalSteps` is a misnomer** (`internal/model/model.go:193`, `engine.go:150`)

   The `TotalSteps` field and its JSON key `total_steps` count *transitions taken*, not *total steps in the workflow* (that's `len(wf.Steps)`). It's used for cycle detection. A name like `TransitionCount` would be clearer, since `TotalSteps` strongly implies "how many steps does this workflow have." The TS interface now also exposes `total_steps`, which the UI could misinterpret.

2. **`WorkflowReviewOutcome` silently skips post-action when worktree is gone** (`internal/session/manager.go:1044`)

   If `wtPath` is empty (worktree was cleaned up, or no matching session found), the post-action is silently skipped. For actions like `merge` this is fine (no worktree needed), but `push` would silently do nothing. Worth at least logging when `result.Action != ""` and `wtPath == ""`.

3. **`ExecutePostAction` partial failure on `push_and_pr`** (`internal/workflow/engine.go:193-199`)

   If push succeeds but PR creation fails, it returns `"pushed"` as the outcome *and* an error. The caller in `executeAndAdvance` checks `err != nil` and treats the whole thing as a failure, which is correct -- but the push has already happened. The workflow will be marked errored, and on retry, push will attempt again (which should be idempotent). This is fine in practice, but the error message could note that the push succeeded to help debugging.

### Minor

4. **`workflow_test.go` doesn't test `Manager.List` or `Manager.Get` failure paths** (`internal/workflow/workflow_test.go`)

   `Get` with a non-existent name, `List` with an empty or non-existent directory -- worth a few small tests for confidence.

5. **`WorkflowPaused` status is defined but never used** (`internal/model/model.go:203`)

   The status constant `WorkflowPaused` exists and the TS type includes `'paused'`, but no code path ever sets a workflow run to paused. Either a future feature (fine, but document it) or dead code.

6. **Workflow handler doesn't validate `outcome` values** (`internal/server/workflow_handlers.go:88`)

   The review endpoint accepts any string as `outcome`. Typos like `"apporve"` would silently fall through to the linear advance path in `ReviewOutcome` rather than matching `on_approve`. Consider validating against known values (`approve`, `changes_requested`).

## Suggestions

- For #1: rename `TotalSteps` to `TransitionCount` (model, store, migration, TS interface).
- For #2: add `log.Printf("workflow: no worktree path for post-action %s, skipping", result.Action)` before the `if` on line 1044.
- For #6: add a quick check in the handler: `if body.Outcome != "approve" && body.Outcome != "changes_requested" { writeError(...); return }`.

## Verdict

**Approve.** The previous review findings have all been addressed in the follow-up commits. The remaining findings are minor or about polish rather than correctness. The architecture is clean -- the engine, parser, and manager have clear boundaries, the tests cover the important state transitions, and the cycle protection is a nice touch.

PR #74, opened 2026-04-13
Dispatch skills should rebase worktree branch onto main before starting work

### LINKED ISSUE #72: Dispatch skills should rebase worktree branch onto main before starting work
Dispatched agents start working on a worktree branch that may be behind main. When the work is done and needs to be merged, conflicts appear that could have been avoided by rebasing first.

This also applies to address-feedback sessions -- the PR branch may have diverged from main since the original implementation.

## Fix

Add a rebase step to the dispatch flow. Two options:

1. **Orchestrator does it** -- in `session.Manager`, before launching the claude process, run `git fetch origin main && git rebase origin/main` in the worktree. If the rebase fails (conflicts), abort and report the error rather than starting a session that will struggle.

2. **Skill instructs the agent** -- add "rebase onto main before starting" to the global_prompt or to each skill template. Simpler but relies on the agent doing it correctly.

Option 1 is better -- it's deterministic, catches conflicts before burning a session, and doesn't depend on agent behaviour.

## Where

`internal/session/manager.go` in the dispatch path, after worktree creation but before launching the claude process. The worktree package already has `UpdateBranch` which does fetch + rebase.

jasonmadigan:
Dispatched agents start working on a worktree branch that may be behind main. When the work is done and needs to be merged, conflicts appear that could have been avoided by rebasing first.

This also applies to address-feedback sessions -- the PR branch may have diverged from main since the original implementation.

## Fix

Add a rebase step to the dispatch flow. Two options:

1. **Orchestrator does it** -- in `session.Manager`, before launching the claude process, run `git fetch origin main && git rebase origin/main` in the worktree. If the rebase fails (conflicts), abort and report the error rather than starting a session that will struggle.

2. **Skill instructs the agent** -- add "rebase onto main before starting" to the global_prompt or to each skill template. Simpler but relies on the agent doing it correctly.

Option 1 is better -- it's deterministic, catches conflicts before burning a session, and doesn't depend on agent behaviour.

## Where

`internal/session/manager.go` in the dispatch path, after worktree creation but before launching the claude process. The worktree package already has `UpdateBranch` which does fetch + rebase.

Closes jasonmadigan/clawdio#72


jasonmadigan (commented):
Same failure on main -- pre-existing, not caused by this PR. Here's the review:

---

## CI Status

No CI checks configured on this branch. The full test suite passes for all PR-touched packages. There's a pre-existing failure in `internal/tmux` (unrelated).

## Summary

Adds a `RebaseOnMain` step to the dispatch flow so worktree branches are rebased onto `origin/main` before launching a Claude session. If the rebase conflicts, it aborts cleanly and returns an error instead of starting a doomed session.

## Findings

**No critical or important issues.**

**Minor:**

1. **`RebaseOnMain` duplicates most of `UpdateBranch`** (`worktree.go:198-219` vs `221-241`). The only difference is the target branch -- `RebaseOnMain` hardcodes `"main"` while `UpdateBranch` takes a branch parameter. `RebaseOnMain` could just be `m.UpdateBranch(wtPath, "main")`. Not a blocker, but it's two copies of the same fetch+rebase+abort logic.

2. **Ordering with `UpdateBranch` for PR items** (`manager.go:150-159`). For PR work items, the flow is: `UpdateBranch(wtPath, prBranch)` then `RebaseOnMain(wtPath)`. This means we fetch twice (once for the PR branch, once for main). The order is correct -- rebase the PR branch tip first, then rebase that onto main -- but worth noting that `RebaseOnMain` after `UpdateBranch` on a PR branch will linearise the PR commits onto main, which changes the branch history. For address-feedback dispatches this is fine (we want the latest main), but it does mean the PR branch in the worktree will diverge from the remote PR branch. The agent will need to force-push if it pushes. This is probably the intended behaviour but worth a brief comment.

3. **`UpdateBranch` failure is logged but not fatal** (`manager.go:152-153`) while `RebaseOnMain` failure is fatal (`manager.go:157-158`). The asymmetry is deliberate (PR branch might not have a remote yet) but could be surprising to future readers.

## Suggestions

- Consider implementing `RebaseOnMain` as `m.UpdateBranch(wtPath, "main")` to avoid duplicating the fetch+rebase+abort pattern.

## Verdict

**Approve.** Clean implementation, good test coverage (happy path + conflict case with abort verification). The duplication with `UpdateBranch` is minor and could be addressed later.

PR #75, opened 2026-04-13
Autopilot mode: auto-dispatch work items based on rules

### LINKED ISSUE #23: Autopilot mode: auto-dispatch work items based on rules
Add an autopilot mode that automatically dispatches incoming work items without manual intervention. Configurable per item type and repo.

Config example:
```json
"autopilot": {
  "enabled": false,
  "rules": [
    {
      "repos": ["jasonmadigan/clawdio"],
      "item_types": ["issue"],
      "skill": "implement-issue",
      "auto_pr": true
    },
    {
      "repos": ["Kuadrant/mcp-gateway"],
      "item_types": ["pr_review"],
      "skill": "review-pr",
      "auto_pr": false
    }
  ]
}
```

When a new work item arrives via GitHub polling and matches a rule:
1. Auto-dispatch with the configured skill
2. If `auto_pr` is true, auto push + create PR when done
3. Items still appear on the board so the user can see what happened

Safety:
- Global on/off toggle in the UI header (visible, obvious)
- Per-repo and per-type scoping so it doesn't run wild
- Rate limiting (max N concurrent auto-dispatched sessions)
- Autonomy rules still apply for external actions
- Clear visual indicator that an item was auto-dispatched vs manual

jasonmadigan:
Add an autopilot mode that automatically dispatches incoming work items without manual intervention. Configurable per item type and repo.

Config example:
```json
"autopilot": {
  "enabled": false,
  "rules": [
    {
      "repos": ["jasonmadigan/clawdio"],
      "item_types": ["issue"],
      "skill": "implement-issue",
      "auto_pr": true
    },
    {
      "repos": ["Kuadrant/mcp-gateway"],
      "item_types": ["pr_review"],
      "skill": "review-pr",
      "auto_pr": false
    }
  ]
}
```

When a new work item arrives via GitHub polling and matches a rule:
1. Auto-dispatch with the configured skill
2. If `auto_pr` is true, auto push + create PR when done
3. Items still appear on the board so the user can see what happened

Safety:
- Global on/off toggle in the UI header (visible, obvious)
- Per-repo and per-type scoping so it doesn't run wild
- Rate limiting (max N concurrent auto-dispatched sessions)
- Autonomy rules still apply for external actions
- Clear visual indicator that an item was auto-dispatched vs manual

Closes jasonmadigan/clawdio#23


jasonmadigan (commented):
Good. I've reviewed everything. Here's my review:

---

## CI Status

Passing (build-and-test, 31s).

## Summary

Adds an autopilot engine that auto-dispatches incoming GitHub work items based on configurable rules (repo + item type scoping). Includes a global toggle in the UI header, rate limiting via max concurrent sessions, auto-PR creation on session completion, and an "auto" badge on cards. Clean separation into `internal/autopilot/` with its own tests.

## Findings

### Important

1. **Struct alignment is broken in `internal/model/model.go:38-41`** -- the `AutopilotDispatched` field uses wider indentation than the fields above it (`ParentID` through `UpdatedAt` are padded with extra spaces/tabs). This looks like tab-vs-space mixing from the diff. The struct should use consistent alignment:

```go
ParentID            string         `json:"parent_id"`
AutopilotDispatched bool           `json:"autopilot_dispatched"`
CreatedAt           time.Time      `json:"created_at"`
UpdatedAt           time.Time      `json:"updated_at"`
```

2. **Auto-PR goroutine in `cmd/clawdio/main.go:83-104` has no error visibility to the user** -- if the push or PR creation fails, it only logs to stdout. The work item stays in whatever state it's in with no indication in the UI that auto-PR failed. Consider updating item status or adding a field so the user can see something went wrong rather than just wondering why no PR appeared.

3. **`ProcessNewItems` doesn't check if an item is already dispatched** (`autopilot.go:72-111`) -- if the poller somehow re-imports the same items (e.g. a race between poll cycles), there's no guard against double-dispatching. The `CreateWorkItem` in the poller likely prevents true duplicates, but worth a defensive check on `item.Status != model.StatusInbox` or `item.AutopilotDispatched`.

### Minor

4. **Blank line at `autopilot_test.go:13-14`** -- extra blank line between imports and the first test function (two blank lines instead of one).

5. **`BroadcastItemUpdate()` called unconditionally after `ProcessNewItems`** (`main.go:237`) -- this fires even when `ProcessNewItems` dispatched nothing (autopilot disabled, no rules matched, at capacity). Not harmful, but a wasted websocket broadcast.

6. **No test for the happy path of `ProcessNewItems`** -- there's a test for the disabled case (`TestProcessNewItemsDisabled`), but no test that verifies items actually get dispatched when enabled with matching rules and a registered skill. The rule matching and toggle logic are well-tested, but the integration path through `ProcessNewItems` with a real store + dispatch function isn't covered for the success case.

## Suggestions

- For finding #2: a lightweight approach would be setting a `pr_url` to something like `"error: push failed"` or adding a status note field. Even just broadcasting an agent event with the error would give the user visibility.
- For finding #3: add `if item.Status != model.StatusInbox { continue }` at the top of the loop as a safety guard.
- Consider adding a happy-path integration test for `ProcessNewItems` that registers a test skill file and asserts the dispatch function is called and the item's status/flags are updated.

## Verdict

**Approve.** The architecture is sound -- clean separation, good use of the callback pattern for dispatch, sensible safety controls (global toggle, rate limiting, scoped rules). The alignment issue (#1) is cosmetic but should be fixed. Finding #2 (silent auto-PR failures) and #3 (no double-dispatch guard) are worth addressing but aren't blocking -- the failure mode is visible in server logs and the poller's dedup logic makes #3 unlikely in practice. Good test coverage on the core matching logic.

jasonmadigan (commented):
All tests pass. Here's the review:

---

## CI Status

Pending (build-and-test still running). Tests pass locally.

## Summary

Adds an autopilot engine that auto-dispatches incoming GitHub work items based on configurable rules (repo + item type scoping). Includes global toggle in the UI header, rate limiting via max concurrent sessions, auto-PR creation on session completion, and an "auto" badge on cards. Previous review feedback (error visibility, double-dispatch guard, happy-path tests) has been addressed in the follow-up commit.

## Findings

### Important

1. **Struct alignment in `internal/model/model.go:37-41` is still inconsistent** -- the previous review flagged this and the fix commit attempted to address it, but the result is that `PRURL`, `ParentID`, `AutopilotDispatched`, `CreatedAt`, and `UpdatedAt` now use wider tab-padding than `ID` through `SkillName` (lines 26-36). The struct has two different alignment zones. Either all fields should align to the widest field, or use minimal alignment throughout.

### Minor

2. **`BroadcastItemUpdate()` called unconditionally after `ProcessNewItems`** (`main.go:237`) -- previous review noted this. Still fires a websocket broadcast even when nothing was dispatched. The conditional `> 0` check is there for the autopilot call, but the first `BroadcastItemUpdate()` at line 233 fires for any new items regardless. Not a bug, just a minor inefficiency.

## Previous Review Feedback -- Status

The prior review raised 6 findings. Checking them against the fix commit (`9172560`):

| # | Finding | Status |
|-|-|-|
| 1 | Struct alignment | Partially fixed -- still inconsistent (see finding #1 above) |
| 2 | Auto-PR error visibility | Fixed -- errors now broadcast as agent events |
| 3 | Double-dispatch guard | Fixed -- `ProcessNewItems` checks `item.Status != StatusInbox \|\| item.AutopilotDispatched` |
| 4 | Extra blank line in test file | Fixed |
| 5 | Unconditional broadcast | Not addressed (minor, acceptable) |
| 6 | No happy-path test | Fixed -- `TestProcessNewItemsHappyPath` and `TestProcessNewItemsSkipsAlreadyDispatched` added |

## Verdict

**Approve.** The architecture is clean, the safety controls are sensible (toggle, rate limiting, scoped rules), and the previous review's important findings have been addressed. The struct alignment issue is cosmetic -- `gofmt` won't catch it since Go aligns struct tags per alignment group, and this struct just has an awkward break at the new field boundary. Not worth blocking on.

Issue #62, opened 2026-04-13
Replace boilerplate README with proper project documentation


jasonmadigan:
The README is still the Vite template boilerplate. Replace with actual project documentation.

## Content

- What clawdio is (one paragraph)
- Screenshot or two of the board
- Quick start (link to Docker for easy route, dev setup for contributors)
- Prerequisites (Go, Node, gh CLI, claude CLI)
- Config walkthrough (config.example.json, what each field does)
- How it works (polling GitHub, dispatching agents, worktrees, skills)
- Link to docs/ for deeper architecture and development guides

## Style

Terse. No walls of text. Code examples over explanations. The docs/ directory already has architecture.md, session-protocol.md, work-items.md, configuration.md, development.md -- the README should link to these, not duplicate them.

PR #88, opened 2026-04-13
Workflow engine: fix fragile state transitions and handover between steps

### LINKED ISSUE #85: Workflow engine: fix fragile state transitions and handover between steps
The implement-and-land workflow is fundamentally fragile. Every step makes assumptions about state that break in practice, causing infinite loops, stuck items, and silent failures.

### Self-approval on own PRs
`gh pr review --approve` and `gh pr review --request-changes` both fail when you're the PR author. Review feedback never reaches GitHub. address-feedback then finds nothing to address, loops forever.

**Partial fix landed:** fall back to posting as a PR comment. But this is a workaround -- the real fix is to either use a bot account for reviews or redesign around the constraint.

### PR number confusion
The workflow used `item.SourceRef` (the issue number, e.g. #62) when it needed the PR number (e.g. #80) for approve/merge operations. Fixed with `PRNumberFromURL(item.PRURL)` but this was a fundamental design oversight.

### Draft PR merge failure
`CreateDraftPR` creates drafts. `MergePR` can't merge drafts. Fixed by adding `MarkPRReady` before merge, but the workflow should decide up front whether to create draft or ready PRs.

### CI timeout
`WaitForChecks` polls CI but the workflow's context expires before CI finishes. Fixed timeout to 10 minutes, but long CI pipelines will still fail. Should decouple CI polling from the workflow context.

### Worktree cleanup race
`CleanupStale` in the poller removes worktrees for merged branches without checking for active sessions. Fixed with an exclusion list, but the cleanup still runs every 30 seconds and can race.

### Stuck sessions
If the server crashes or tmux fails, sessions stay "running" forever with no process. Fixed with a 30-second health check, but the root cause (no transactional session lifecycle) remains.

### Infinite review loop
review-pr requests changes -> feedback never posted to GitHub -> address-feedback finds nothing -> pushes nothing -> loops back to review -> same findings -> loop forever. Only breaks when max_review_rounds (3) is hit.

### 1. Make every external action idempotent and observable
Every action (push, create PR, post review, merge) should:
- Check the current state before acting (does PR exist? is it draft? is CI passing?)
- Be retryable without side effects
- Log what it did and didn't do

### 2. Post all review feedback to GitHub unconditionally
Don't use `gh pr review` at all for self-authored PRs. Always use `gh pr comment` or `gh api` to post review text. The review-pr agent writes the review; the orchestrator posts it as a comment with a structured prefix (e.g. `[clawdio review]`) so address-feedback can find it.

### 3. Decouple CI polling from workflow context
CI check should be a separate async step, not blocking inside the merge post-action. The workflow should: mark the item as "waiting for CI", set a timer, check back. If CI fails, pause the workflow with a clear error.

### 4. State machine with explicit transitions
Replace the current implicit state flow with an explicit state machine:
```
idle -> implementing -> pushing -> pr_created -> reviewing -> 
  approved -> ci_checking -> merging -> done
  changes_requested -> addressing_feedback -> pushing -> reviewing
```
Each transition validates preconditions and handles failures.

### 5. Clean up the PR lifecycle
- Decide: draft or ready PRs? (ready makes more sense for autopilot)
- Store PR state on the work item (draft, ready, CI status, review status)
- Don't re-create PRs that already exist
- Don't re-push when nothing changed

jasonmadigan:
The implement-and-land workflow is fundamentally fragile. Every step makes assumptions about state that break in practice, causing infinite loops, stuck items, and silent failures.

### Self-approval on own PRs
\`gh pr review --approve\` and \`gh pr review --request-changes\` both fail when you're the PR author. Review feedback never reaches GitHub. address-feedback then finds nothing to address, loops forever.

**Partial fix landed:** fall back to posting as a PR comment. But this is a workaround -- the real fix is to either use a bot account for reviews or redesign around the constraint.

### PR number confusion
The workflow used \`item.SourceRef\` (the issue number, e.g. #62) when it needed the PR number (e.g. #80) for approve/merge operations. Fixed with \`PRNumberFromURL(item.PRURL)\` but this was a fundamental design oversight.

### Draft PR merge failure
\`CreateDraftPR\` creates drafts. \`MergePR\` can't merge drafts. Fixed by adding \`MarkPRReady\` before merge, but the workflow should decide up front whether to create draft or ready PRs.

### CI timeout
\`WaitForChecks\` polls CI but the workflow's context expires before CI finishes. Fixed timeout to 10 minutes, but long CI pipelines will still fail. Should decouple CI polling from the workflow context.

### Worktree cleanup race
\`CleanupStale\` in the poller removes worktrees for merged branches without checking for active sessions. Fixed with an exclusion list, but the cleanup still runs every 30 seconds and can race.

### Stuck sessions
If the server crashes or tmux fails, sessions stay "running" forever with no process. Fixed with a 30-second health check, but the root cause (no transactional session lifecycle) remains.

### Infinite review loop
review-pr requests changes -> feedback never posted to GitHub -> address-feedback finds nothing -> pushes nothing -> loops back to review -> same findings -> loop forever. Only breaks when max_review_rounds (3) is hit.

### 1. Make every external action idempotent and observable
Every action (push, create PR, post review, merge) should:
- Check the current state before acting (does PR exist? is it draft? is CI passing?)
- Be retryable without side effects
- Log what it did and didn't do

### 2. Post all review feedback to GitHub unconditionally
Don't use \`gh pr review\` at all for self-authored PRs. Always use \`gh pr comment\` or \`gh api\` to post review text. The review-pr agent writes the review; the orchestrator posts it as a comment with a structured prefix (e.g. \`[clawdio review]\`) so address-feedback can find it.

### 3. Decouple CI polling from workflow context
CI check should be a separate async step, not blocking inside the merge post-action. The workflow should: mark the item as "waiting for CI", set a timer, check back. If CI fails, pause the workflow with a clear error.

### 4. State machine with explicit transitions
Replace the current implicit state flow with an explicit state machine:
\`\`\`
idle -> implementing -> pushing -> pr_created -> reviewing -> 
  approved -> ci_checking -> merging -> done
  changes_requested -> addressing_feedback -> pushing -> reviewing
\`\`\`
Each transition validates preconditions and handles failures.

### 5. Clean up the PR lifecycle
- Decide: draft or ready PRs? (ready makes more sense for autopilot)
- Store PR state on the work item (draft, ready, CI status, review status)
- Don't re-create PRs that already exist
- Don't re-push when nothing changed

Closes jasonmadigan/clawdio#85


jasonmadigan (commented):
Confirmed -- `WaitForChecks` has its own internal `deadline` from the `timeout` param, plus it respects `ctx` cancellation. So the `ciCtx` with a 10-minute timeout and the 10-minute `timeout` parameter are redundant. Not a bug, but worth noting.

I have enough to write the review.

---

## CI Status

**Pending** -- `build-and-test` is still running. Cannot confirm the build passes.

## Summary

Adds explicit workflow phase tracking to the engine, makes post-actions (push, PR creation, merge) idempotent by checking preconditions before acting, and surfaces phase/error information in the UI. Also extracts the review-posting logic into `PostReviewToGitHub` with a `[clawdio review]` prefix for structured discovery.

## Findings

### Important

1. **`PostReviewToGitHub` posts fallback comments to the wrong target** (`engine.go:377,386`). When self-review fails, it calls `github.PostComment(ctx, item.SourceRef, prefixed)`. But `item.SourceRef` is the *source issue* (e.g. `repo#62`), not the PR (e.g. `repo#80`). The `prNumber` is right there in the function arguments. The fallback comment should be posted to the PR, not the originating issue. This is the exact bug class the PR description calls out ("PR number confusion"), but it's been reintroduced in the extracted function.

2. **`PhasePRCreated` is defined but never set** (`model.go:201`). The phase constant exists, yet no code path ever transitions a run into this phase. After `executePushAndPR` succeeds, the phase remains `pushing` -- the caller (`executeAndAdvance` in `manager.go`) just calls `PostActionDone` which advances to the next step. Either remove the unused constant or set it after PR creation in `executePushAndPR`.

3. **Missing CSS for new UI classes** (`Panel.tsx:52,55`). `.workflow-phase` and `.workflow-error` are rendered but have no corresponding CSS. They'll appear as unstyled text.

### Minor

4. **Redundant CI timeout** (`engine.go:314-317`). `ciCtx` has a 10-minute timeout, and `WaitForChecks` is also passed a 10-minute timeout (it manages its own internal deadline). The `ciCtx` is fine as a safety net, but the duplication is confusing -- consider dropping the `timeout` param and just relying on context cancellation, or vice versa.

5. **`executeMerge` reaches into store for phase updates** (`engine.go:306-311, 335-340`). The post-action functions are otherwise clean -- they do their work and return an outcome. But `executeMerge` directly fetches the active workflow run and updates its phase. This breaks the separation between the engine (which owns phase transitions) and the post-action functions (which execute external operations). The phase updates should live in the caller, not inside `executeMerge`.

6. **`phaseForSkill` defaults to `PhaseImplementing`** (`engine.go:424`). Any unrecognised skill silently maps to "implementing". A custom skill like `"run-tests"` would show as "Implementing" in the UI. Consider returning `PhaseIdle` or the skill name itself for unknowns.

## Suggestions

- Fix #1 by using `prNumber` (or constructing a source ref from `item.PRURL`) in the `PostComment` fallback calls.
- For #2, either use `PhasePRCreated` at the right moment or remove it to avoid dead constants.
- Add basic CSS for `.workflow-phase` and `.workflow-error`.

## Verdict

**Request changes.** The fallback comment target bug (#1) means review feedback on self-authored PRs will still land on the wrong issue, which was one of the core problems this PR is fixing. The rest is solid work -- the idempotent actions, phase tracking, and test coverage are good improvements.

jasonmadigan (commented):
The call site guards with `item.PRURL != ""`, and if a PR URL exists, `SourceRef` will also be set (GitHub-sourced items always have both). So the empty `SourceRef` edge case isn't practically reachable here. Acceptable.

---

## CI Status

**No checks reported** on this branch. Cannot confirm the build passes.

## Summary

Adds explicit workflow phase tracking to the engine, makes post-actions (push, PR creation, merge) idempotent by checking preconditions before acting, and surfaces phase/error information in the UI. Extracts review-posting logic into `PostReviewToGitHub` with a `[clawdio review]` prefix for structured discovery. A follow-up commit addresses all six findings from the previous review.

## Previous review

A review was posted as a comment on 2026-04-13. It identified six issues (fallback comment target, unused `PhasePRCreated` constant, missing CSS, redundant CI timeout, `executeMerge` reaching into store, `phaseForSkill` default). All six have been addressed in commit `e96586e`:

1. Fallback comment target -- fixed, now constructs `prRef` from `repo + "#" + prNumber`
2. `PhasePRCreated` -- removed from Go constants
3. Missing CSS -- `.workflow-phase` and `.workflow-error` styles added
4. Redundant CI timeout -- removed the extra `ciCtx`, relies on `WaitForChecks`'s own deadline
5. `executeMerge` reaching into store -- replaced with `PhaseCallback` pattern, phase transitions now owned by the caller in `manager.go`
6. `phaseForSkill` default -- changed to `PhaseIdle`

## Findings

### Minor

1. **`pr_created` phase in TypeScript types but no Go constant** (`types.ts:112`, `Panel.tsx:29`). The Go constant `PhasePRCreated` was removed, but the TypeScript union type and `PHASE_LABELS` map still include `pr_created`. Nothing sets this phase, so the label is dead code. Remove it from both files, or add a phase transition for it if it's intended to be used.

2. **`HasNewCommits` relies on remote tracking** (`github.go:270`). It uses `git log HEAD --not --remotes` which compares against all remote refs. If the remote tracking info is stale (no recent fetch), this could report commits as "new" when they've already been pushed, or miss the case where the remote was force-pushed. In the worktree context this is likely fine since the worktree was just cloned, but worth noting.

3. **`executePushAndPR` creates ready PRs but doesn't update `PRDraft`** (`engine.go:232-265`). The function calls `st.UpdateWorkItemPRState(item.ID, false, "", "")` which correctly sets `pr_draft = false`, but it also blanks out `pr_ci_status` and `pr_review_status` even if the PR already existed. For existing PRs, these fields might already have meaningful values from a prior cycle.

## Verdict

**Approve.** The previous review's findings are all addressed. The remaining items are minor -- no bugs, no logic errors, no missing edge case handling. The idempotent actions, phase tracking, and test coverage are solid improvements. Good to merge once CI is confirmed passing.

PR #102, opened 2026-04-13
Contract output parser returns wrong verdict despite correct text in summary

### LINKED ISSUE #101: Contract output parser returns wrong verdict despite correct text in summary
## Bug

Session summary contains \`VERDICT: approve\` but contract output parser returned \`map[verdict:changes_requested]\`.

## Evidence

Session 5a8e787f summary ends with:
\`\`\`
Worth fixing but not blocking.

VERDICT: approve
\`\`\`

But the log shows:
\`\`\`
workflow 3926d344: contract outputs for review-pr: map[verdict:changes_requested]
\`\`\`

The regex \`VERDICT:\\s*(approve|changes_requested)\` matches correctly in isolation (tested). The summary stored in the DB contains the right text. Something between the DB read and the parse is wrong.

## Possible causes

1. **Timing** -- the summary may not have been written to DB when \`advanceWithContract\` read it. \`finish()\` calls \`UpdateSessionStatus\` then \`GetSession\`, but there could be a race.
2. **seenLines dedup** -- the line-level dedup added for pipe-pane may be filtering out the result event, causing \`lastText\` to not include the VERDICT line. The result event replaces \`lastText\` (line 718), but if the event was deduplicated, \`lastText\` stays as the previous text block.
3. **Contract not loaded** -- the skill contract might not have been parsed from the frontmatter, causing fallback to the default value (\`changes_requested\`).

## Impact

This causes autopilot to dispatch address-feedback when it should merge, wasting time and tokens. Critical for Ship It reliability.

jasonmadigan:
## Bug

Session summary contains \`VERDICT: approve\` but contract output parser returned \`map[verdict:changes_requested]\`.

## Evidence

Session 5a8e787f summary ends with:
\`\`\`
Worth fixing but not blocking.

VERDICT: approve
\`\`\`

But the log shows:
\`\`\`
workflow 3926d344: contract outputs for review-pr: map[verdict:changes_requested]
\`\`\`

The regex \`VERDICT:\\s*(approve|changes_requested)\` matches correctly in isolation (tested). The summary stored in the DB contains the right text. Something between the DB read and the parse is wrong.

## Possible causes

1. **Timing** -- the summary may not have been written to DB when \`advanceWithContract\` read it. \`finish()\` calls \`UpdateSessionStatus\` then \`GetSession\`, but there could be a race.
2. **seenLines dedup** -- the line-level dedup added for pipe-pane may be filtering out the result event, causing \`lastText\` to not include the VERDICT line. The result event replaces \`lastText\` (line 718), but if the event was deduplicated, \`lastText\` stays as the previous text block.
3. **Contract not loaded** -- the skill contract might not have been parsed from the frontmatter, causing fallback to the default value (\`changes_requested\`).

## Impact

This causes autopilot to dispatch address-feedback when it should merge, wasting time and tokens. Critical for Ship It reliability.

Closes jasonmadigan/clawdio#101


jasonmadigan (commented):
All passing. Here's the review:

---

## CI Status

All checks passing.

## Summary

Fixes a bug where the contract output parser returned `changes_requested` despite the session summary containing `VERDICT: approve`. Two root causes addressed: (1) `result` events from `stream-json` could be silently dropped by the `seenLines` deduplication map, meaning `lastText` never got updated with the final verdict, and (2) `finish()` used `sess.Summary` from a DB round-trip instead of the `summary` argument already in hand, introducing a race window.

## Findings

**Minor:**

1. `internal/skill/output_test.go:178` -- `string(rune('A'+i))` with `i` going up to 29 will produce non-letter characters for `i >= 26` (brackets, backslash, etc.). Works fine for the test purpose since they're just filler lines, but it reads as though it's meant to produce `A` through `Z` then wraps oddly. Cosmetic only.

2. `internal/session/manager.go:698-703` -- the `result` exemption means a genuinely duplicated result event will be processed twice. In practice this just re-assigns `*lastText` to the same value, so it's harmless. But worth noting that `seenLines` now grows unboundedly for `result` events that aren't added to the map. This is fine since there's typically one result event per session, but the comment could mention that.

## Suggestions

None worth acting on. Both fixes are correct and minimal. The tests cover the exact regression scenario plus the inverse (non-result dedup still works). The `output_test.go` addition confirms the parser itself wasn't the problem, which is good defensive coverage.

## Verdict

Clean fix, well-scoped, good test coverage. Both root causes are real -- the `seenLines` dedup is the more likely culprit in practice, and the `sess.Summary` override is a good belt-and-braces addition against the DB race.

VERDICT: approve

PR #122, opened 2026-04-14
Dispatch API should accept skill override in request body

### LINKED ISSUE #120: Dispatch API should accept skill override in request body
The dispatch endpoint (\`POST /api/work-items/{id}/dispatch\`) ignores skill_name in the request body. To dispatch with a specific skill, you have to PATCH the work item first then dispatch -- two calls for one action.

The UI works around this (the dispatch modal patches the item's skill_name before calling dispatch), but the API should support it directly:

\`\`\`json
POST /api/work-items/{id}/dispatch
{"skill_name": "review-pr", "extra_prompt": "..."}
\`\`\`

If \`skill_name\` is provided in the body, use it for this dispatch without permanently changing the work item's skill_name field.

jasonmadigan:
The dispatch endpoint (\`POST /api/work-items/{id}/dispatch\`) ignores skill_name in the request body. To dispatch with a specific skill, you have to PATCH the work item first then dispatch -- two calls for one action.

The UI works around this (the dispatch modal patches the item's skill_name before calling dispatch), but the API should support it directly:

\`\`\`json
POST /api/work-items/{id}/dispatch
{"skill_name": "review-pr", "extra_prompt": "..."}
\`\`\`

If \`skill_name\` is provided in the body, use it for this dispatch without permanently changing the work item's skill_name field.

Closes jasonmadigan/clawdio#120


jasonmadigan:
[clawdio review]

---

## CI Status

Passing (build-and-test).

## Summary

Adds `skill_name` to the dispatch request body struct. When provided, it's used for skill resolution via a shallow copy of the work item, without persisting the override to the database.

## Findings

No issues found. The implementation is clean:

- `resolveItem := item` is a safe value copy since `GetWorkItem` returns `model.WorkItem` by value and `Resolve` takes it by value. The only non-scalar field (`Labels []string`) isn't mutated, so no aliasing risk.
- The original `item` (without the override) is passed to `Dispatch`, so the override is truly transient.
- The comment on line 157 accurately describes the behaviour.
- Removed comments were noise (the code is self-explanatory).

## Suggestions

- **Minor:** No test coverage for the new skill override path in the dispatch handler. An integration test confirming that `skill_name` in the body resolves the correct skill without modifying the stored work item would be valuable, but not blocking for a change this small.

## Verdict

Approve. Small, focused change that does exactly what the issue asks for. The copy-based approach avoids side effects cleanly.

```
VERDICT: approve
```

Issue #115, opened 2026-04-14
Docker quickstart: end-user getting started guide


jasonmadigan:
With the Dockerfile (#64) and embedded frontend (#63) merged, we can offer a much simpler onboarding for users who just want to USE clawdio, not develop it.

## Target audience

Someone who has Docker, a GitHub account, and a Claude/Anthropic API key. They don't need Go, Node, or air installed. They don't need to clone the repo.

## Quickstart

Should be this simple:

\`\`\`bash
docker run -it \\
  -e ANTHROPIC_API_KEY=sk-... \\
  -v ~/.config/gh:/home/clawdio/.config/gh:ro \\
  -v clawdio-data:/data \\
  -p 8077:8077 \\
  ghcr.io/jasonmadigan/clawdio:latest
\`\`\`

Open http://localhost:8077. Configure repos in Settings. Start dispatching.

## What this needs

- **CI to build and push the Docker image** to ghcr.io on tag/release (or on main merge)
- **README update** with a "Quick start (Docker)" section before the dev setup section
- **First-run experience** -- if no config.json exists, the UI should guide through setup (add repos, set poll queries) rather than showing an empty board
- **gh auth** -- see #115 for authentication in containers

## Relates to

- #63 (embedded frontend) -- merged
- #64 (Dockerfile) -- merged
- #62 (README) -- merged, needs Docker section added
- #115 (gh auth in containers)

Issue #83, opened 2026-04-13
Improve logging: structured output, workflow lifecycle, error context


jasonmadigan:
Logging is currently a mix of ad-hoc Printf calls with inconsistent formatting. Hard to follow what's happening during autopilot workflows, session lifecycle, and error scenarios.

## Problems

- **No structure** -- plain text lines with inconsistent prefixes (sometimes "autopilot:", sometimes "workflow:", sometimes "session %s:")
- **Workflow lifecycle is hard to follow** -- individual events log but there's no summary of "item X: implement -> push -> PR #80 -> review -> approve -> merge" as a cohesive flow
- **Errors lack context** -- "CI check: context deadline exceeded" doesn't say which item, which PR, which workflow run
- **No log levels** -- everything is Printf. Can't filter debug noise from important events
- **Session events are noisy** -- "event type=assistant" on every stream-json line floods the log. Useful for debugging, useless in normal operation
- **Poller logs are repetitive** -- "skipping cleanup of X (active session)" every 30 seconds
- **No request ID tracing** -- can't correlate an API request through dispatch -> session -> post-action -> workflow

## What good looks like

**Structured log lines** with consistent fields:
\`\`\`
level=info component=autopilot item=clawdio#62 action=dispatched skill=implement-issue
level=info component=workflow run=95b5975c step=2/3 action=review-pr status=completed verdict=approve
level=info component=workflow run=95b5975c step=3/3 action=merge pr=#80 status=success
level=error component=workflow run=e1b01684 step=3/3 action=merge error="CI checks failed"
level=debug component=session id=c952ec55 event=assistant
\`\`\`

**Workflow summary on completion:**
\`\`\`
level=info component=autopilot item=clawdio#62 workflow=implement-and-land result=merged pr=#80 duration=3m42s
\`\`\`

## Scope

- **Session manager** -- reduce event-level noise to debug, keep dispatch/complete/error at info
- **Autopilot engine** -- log rule matches, dispatch decisions, capacity state
- **Workflow engine** -- log step transitions, post-actions, review verdicts, final outcome with duration
- **Poller** -- reduce repetitive "skipping cleanup" to debug, keep import/sync at info
- **Server** -- request logging with method, path, duration, status code
- **Health check** -- log when stuck sessions are detected and cleaned up

## Implementation

Could use \`log/slog\` (stdlib since Go 1.21) for structured logging. Or keep Printf but with a consistent format. slog is cleaner but a bigger change.

[...]

PR #124, opened 2026-04-14
Health check should detect and recover stuck workflow runs, not just stuck sessions

### LINKED ISSUE #119: Health check should detect and recover stuck workflow runs, not just stuck sessions
The session health check catches stuck sessions (running with no process) but doesn't handle stuck workflow runs. When a session errors mid-workflow, the workflow run stays in "running" state with no active session and never recovers.

## Observed

1. Ship It dispatched on #115
2. implement-issue completed, address-feedback completed
3. review-pr session errored ("server restarted")
4. Session health check caught the stuck session and marked it errored
5. Work item moved to "ready" (session error handling)
6. But the workflow run stayed at status=running, phase=reviewing
7. No new session dispatched, workflow stuck forever

## Root cause

`CleanupOrphanedSessions` marks stuck sessions as errored but doesn't call `advanceWorkflow`. The normal `finish()` path triggers workflow advancement on session error (the engine sets the run to errored and pauses). But the orphan cleanup path skips this -- it just marks sessions in the DB and moves on. The workflow run never learns its session died.

Fix: after `CleanupOrphanedSessions` runs, iterate the cleaned sessions and call `advanceWorkflow` for any that have a `workflow_run_id`. Or: the health check should do this when it finds stuck workflow runs with no active sessions.

## Fix

The health check should also scan workflow runs:
- Find workflow runs with status=running
- Check if they have an active session
- If not, and no session has been dispatched recently, either:
  - Retry the current step (dispatch a new session for the current skill)
  - Or pause the workflow with an error

## Relates to

- #85 (state machine, merged) -- workflow recovery is a state machine concern
- #83 (logging, merged) -- stuck workflows need clear logging

jasonmadigan:
The session health check catches stuck sessions (running with no process) but doesn't handle stuck workflow runs. When a session errors mid-workflow, the workflow run stays in "running" state with no active session and never recovers.

## Observed

1. Ship It dispatched on #115
2. implement-issue completed, address-feedback completed
3. review-pr session errored ("server restarted")
4. Session health check caught the stuck session and marked it errored
5. Work item moved to "ready" (session error handling)
6. But the workflow run stayed at status=running, phase=reviewing
7. No new session dispatched, workflow stuck forever

## Fix

The health check should also scan workflow runs:
- Find workflow runs with status=running
- Check if they have an active session
- If not, and no session has been dispatched recently, either:
  - Retry the current step (dispatch a new session for the current skill)
  - Or pause the workflow with an error

## Relates to

- #85 (state machine) -- workflow recovery is a state machine concern
- #83 (logging) -- stuck workflows need clear logging

Closes jasonmadigan/clawdio#119


jasonmadigan:
[clawdio review]

I've completed the review. Here's my assessment:

---

## CI Status

Passing (build-and-test: 44s).

## Summary

Adds workflow run recovery to the existing session health check. When a workflow run is stuck in `running` state with no active session, the health check now retries the current step by re-dispatching it. Also adds a startup cleanup for orphaned workflow runs (marking them errored), mirroring the existing `CleanupOrphanedSessions` pattern. Tests cover the store-level queries well.

## Findings

### Important

1. **Race condition: no grace period on stuck detection** (`internal/store/store.go:622-645`)

   `StuckWorkflowRuns()` has no time-based guard. In `DispatchWorkflow` (manager.go:1003-1016), there's a window between `CreateWorkflowRun` (status=running) and `dispatchWorkflowStep` (which creates the session). If the 30-second health check fires during that gap, the run will be falsely detected as stuck and retried -- potentially double-dispatching.

   The existing session cleanup (`cleanupStuckSessions`) avoids this via the `m.IsRunning()` in-memory check. The workflow recovery has no equivalent guard. Adding something like `AND wr.updated_at < datetime('now', '-60 seconds')` to the query would prevent false positives on recently-created or recently-advanced runs.

2. **Retry has no attempt limit** (`internal/session/manager.go:180-244`)

   If `dispatchWorkflowStep` succeeds but the session immediately fails again (e.g. a broken skill, bad worktree), the next health check tick (30s later) will detect it as stuck again and retry. This creates an infinite retry loop. The issue description mentions "retry or pause" -- currently there's no mechanism to pause after repeated failures. A retry counter on the workflow run or a check against the last session's age would prevent this.

### Minor

3. **Dead code: `LastSessionForWorkflowRun`** (`internal/store/store.go:648-660`)

   Defined and tested but never called anywhere outside tests. The recovery code doesn't use it. Either use it (e.g. to implement the grace period or retry limit checks suggested above) or drop it to avoid confusion.

4. **Parallel step recovery not handled** (`internal/session/manager.go:206-211`)

   The recovery bails when `step.Skill == ""` (parallel steps have no single skill). This is logged and errored, which is fine for now, but worth a note that parallel step recovery is a known gap.

## Suggestions

- Add a time guard to `StuckWorkflowRuns` -- e.g. only return runs whose `updated_at` is older than 60 seconds. This matches the intent described in the issue ("no session has been dispatched recently") and prevents the race.
- Add a retry limit. Could be as simple as checking how many errored sessions already exist for this workflow run, and marking it errored instead of retrying after N attempts (2-3 seems reasonable).
- Remove `LastSessionForWorkflowRun` unless it's intended for one of the above improvements -- in which case, wire it up.

## Verdict

The core approach is sound -- detecting stuck workflow runs and retrying is exactly what the issue asks for, and the store queries and tests are well done. The race condition and unbounded retry are real problems that should be addressed before merging.

```
VERDICT: changes_requested
```

jasonmadigan:
[clawdio review]

Good, CI is now passing. Here's my review:

---

## CI Status

Passing (build-and-test: 38s).

## Summary

Adds workflow run recovery to the health check loop and startup cleanup for orphaned workflow runs. When a workflow run is stuck in `running` state with no active session, the health check retries the current step (up to 3 attempts) or marks the run errored. A previous review raised three issues (race condition, unbounded retries, dead code) -- the follow-up commit addresses all three.

## Previous Review Feedback

The first review requested changes for:

1. **Race condition / grace period** -- Fixed. `StuckWorkflowRuns()` now includes `AND wr.updated_at < datetime('now', '-60 seconds')`. The test backdates rows past the grace period and verifies a recent run is excluded.
2. **Unbounded retries** -- Fixed. `ErroredSessionsForWorkflowRun` counts prior failures and `maxWorkflowStepRetries` (3) caps retries before marking the run errored.
3. **Dead code (`LastSessionForWorkflowRun`)** -- Fixed. Replaced with `ErroredSessionsForWorkflowRun` which is wired into the recovery logic. Tests updated accordingly.

All three issues addressed cleanly.

## Findings

### Minor

1. **`ErroredSessionsForWorkflowRun` counts all errored sessions for the run, not per-step** (`internal/store/store.go:652-658`)

   If step 1 produced an errored session, then step 2 gets stuck, the retry counter starts at 1 instead of 0. With `maxWorkflowStepRetries = 3` this means step 2 only gets 2 retries. For a practical retry limit of 3, this is probably fine -- a run that's accumulated 3 errored sessions across any steps is worth stopping. But worth noting the semantics: it's "total errored sessions per run" not "retries per step".

2. **`UpdateWorkflowRun` errors silently dropped** (`internal/session/manager.go:207,217,226,237,247,260`)

   Every `m.store.UpdateWorkflowRun(run)` call in the recovery function ignores the returned error. This matches the existing codebase pattern (the same is true across `engine.go` and the rest of `manager.go`), so it's consistent. Not a blocking issue, but if `UpdateWorkflowRun` fails during recovery, the run stays stuck and will be retried next tick -- which is actually reasonable self-healing behaviour.

## Suggestions

None blocking. The implementation is solid and the review feedback was addressed well.

## Verdict

The approach is sound, the grace period and retry limit address the real risks, tests cover the key scenarios (no sessions, active session, only errored sessions, recent run, completed run), and CI passes. The minor findings above are informational, not blocking.

```
VERDICT: approve
```

Issue #109, opened 2026-04-14
Poller: reopened issues should return to inbox


jasonmadigan:
When an issue is closed, the poller moves the work item to done. But when the same issue is reopened on GitHub, the work item stays at done. It should move back to inbox.

The poller's SyncStatuses handles closed/merged -> done but has no path for reopen -> inbox.

Fix: in the sync loop, if a work item is done but the upstream issue/PR is open, move it back to inbox.

Issue #100, opened 2026-04-13
Ship It should resume from existing state, not restart from scratch


jasonmadigan:
When Ship It is triggered on an item that already has a PR, previous sessions, or a partially completed workflow, it starts from the beginning (implement-issue) instead of resuming from the current state.

## Observed behaviour

1. Ship It dispatched on issue #25
2. Implement-issue completed, pushed, created PR #99
3. Review-pr ran, requested changes
4. Workflow errored on unknown post-action (separate bug, fixed)
5. User reset item to inbox and clicked Ship It again
6. Ship It started implement-issue from scratch in a new worktree
7. Previous PR #99 with all the implementation work sits unused

## Expected behaviour

Ship It should assess the current state before deciding where to start:

- **PR already exists?** Skip implement and push. Start at review.
- **Review feedback exists?** Start at address-feedback.
- **CI passing?** Skip to merge.
- **Already merged?** Mark done.

## Implementation

Before dispatching the first workflow step, check:

1. Does the work item have a pr_url? If so, check if the PR is open/merged/closed.
2. Are there previous completed sessions? What was the last skill and outcome?
3. Does the worktree from the previous run still exist? Reuse it if so.

Map the current state to the right workflow step and start there:

| State | Start at |
|-|-|
| No PR, no sessions | implement-issue (beginning) |
| PR exists, no review | review-pr |
| PR exists, review requested changes | address-feedback |
| PR exists, review approved, CI pending | check_ci |
| PR exists, CI green | merge |
| PR merged | done |

## Relates to

- #85 (state machine, merged) -- the state machine should be resumable from any state
- #77 (per-item autopilot, merged) -- Ship It is the trigger for this

Issue #86, opened 2026-04-13
Architecture: decouple from GitHub with Source, Workspace, and Target abstractions


jasonmadigan:
## The problem

GitHub is hardcoded as both the source of work and the destination of results. Every layer assumes GitHub: the poller, the post-actions, the workflow engine, the status sync, the PR lifecycle, even the skill prompts. Adding Jira (or Linear, or GitLab, or plain file-based work) would require changes across the entire codebase.

This isn't a "clean up the GitHub code" issue. It's a fundamental abstraction gap. The system conflates three independent concerns that need separate interfaces.

### 1. Source -- where work comes from

A Source ingests work items and receives status updates. It knows how to poll, how to parse items, and how to sync state back.

```go
type Source interface {
    // Poll returns new and updated work items.
    Poll(ctx context.Context) ([]model.WorkItem, error)
    
    // SyncStatus pushes clawdio's state back to the source.
    // e.g. add labels, transition Jira tickets, post comments.
    SyncStatus(ctx context.Context, item model.WorkItem, status model.WorkItemStatus) error
    
    // PostComment adds a comment/note to the source item.
    PostComment(ctx context.Context, item model.WorkItem, body string) error
}
```

Implementations:
- `GitHubSource` -- polls via `gh` CLI, syncs via labels/comments (what we have now)
- `JiraSource` -- polls via Jira API, syncs via ticket transitions and comments
- `LinearSource` -- polls via Linear API
- `ManualSource` -- items created via the UI (already exists)
- `WebhookSource` -- items pushed via HTTP webhook

A work item's `source` field already tracks where it came from. The source adapter is resolved from this field.

### 2. Workspace -- where work happens

A Workspace provides an isolated environment for the agent. It manages creation, cleanup, and the agent's working directory.

```go
type Workspace interface {
    // Create sets up an isolated workspace for a work item.
    // Returns the path the agent should work in.
    Create(ctx context.Context, item model.WorkItem) (string, error)
    
    // Cleanup removes the workspace after work is done.
    Cleanup(ctx context.Context, path string) error
    
    // IsClean returns true if the workspace has no uncommitted changes.
    IsClean(ctx context.Context, path string) bool
    
    // ActiveWorkspaces returns paths that have running sessions (for cleanup protection).
    ActiveWorkspaces() []string
}
```

Implementations:
- `GitWorktreeWorkspace` -- creates git worktrees under `~/.clawdio/worktrees/` (what we have now)
[...]

PR #129, opened 2026-04-14
DB as source of truth: own all state locally, sync to/from external sources

### LINKED ISSUE #110: DB as source of truth: own all state locally, sync to/from external sources
## Problem

Clawdio currently relies on GitHub for state in several places:
- Poller checks GitHub to determine if items are open/closed/merged
- Status sync pushes labels to GitHub as the canonical indicator of progress
- PR state (draft, ready, CI status) is queried from GitHub on demand
- Reopening an issue on GitHub doesn't update local state (#109)

This breaks when:
- GitHub goes down (frequently)
- The source isn't GitHub (Jira, Slack, manual items)
- Network is unavailable (offline/containerised)

## Principle

**The SQLite database is the single source of truth for all work item state.** External sources (GitHub, Jira) feed state IN and receive status updates OUT, but the DB is authoritative.

If GitHub is down, clawdio should still:
- Show all work items and their current state
- Dispatch agents and run workflows
- Track session output and workflow progress
- Queue status sync for when connectivity returns

## What changes

**Work item lifecycle is DB-driven:**
- States (inbox, ready, in_progress, needs_review, done) are owned by the DB
- Transitions happen in the DB first, then get synced to sources
- External state changes (issue closed on GitHub) are sync events that propose DB updates, not direct mutations

**Source sync becomes bidirectional and resilient:**
- Inbound: source proposes state changes (issue closed -> suggest moving to done)
- Outbound: DB state changes get pushed to sources (moved to in_progress -> add label)
- Failed syncs are queued and retried, not dropped
- Sync queue persisted in SQLite so it survives restarts

**PR/review state tracked locally:**
- pr_state (draft, ready, merged, closed) stored on the work item
- ci_status (pending, passing, failing) stored on the work item
- review_verdict (approve, changes_requested) stored on the work item
- Updated by sync, used by the workflow engine without querying GitHub

## Schema additions

```sql
ALTER TABLE work_items ADD COLUMN pr_state TEXT DEFAULT '';
ALTER TABLE work_items ADD COLUMN ci_status TEXT DEFAULT '';
ALTER TABLE work_items ADD COLUMN review_verdict TEXT DEFAULT '';

CREATE TABLE sync_queue (
    id TEXT PRIMARY KEY,
    work_item_id TEXT NOT NULL,
    target TEXT NOT NULL,         -- github, jira, etc.
    action TEXT NOT NULL,         -- add_label, post_comment, transition
    payload TEXT NOT NULL,        -- JSON
    status TEXT DEFAULT 'pending', -- pending, synced, failed
    retry_count INTEGER DEFAULT 0,
    created_at DATETIME DEFAULT CURRENT_TIMESTAMP,
    synced_at DATETIME
);
```

## Impact on other issues

- ~~#86 (Source/Workspace/Target)~~ -- merged, Source.SyncStatus becomes async, writes to sync_queue
- #109 (reopen -> inbox) -- trivial once state is DB-driven
- ~~#101 (parser bug)~~ -- fixed, verdict stored in DB, not re-queried from GitHub
- #100 (Ship It resume) -- all state needed to resume is in the DB

jasonmadigan:
## Problem

Clawdio currently relies on GitHub for state in several places:
- Poller checks GitHub to determine if items are open/closed/merged
- Status sync pushes labels to GitHub as the canonical indicator of progress
- PR state (draft, ready, CI status) is queried from GitHub on demand
- Reopening an issue on GitHub doesn't update local state (#109)

This breaks when:
- GitHub goes down (frequently)
- The source isn't GitHub (Jira, Slack, manual items)
- Network is unavailable (offline/containerised)

## Principle

**The SQLite database is the single source of truth for all work item state.** External sources (GitHub, Jira) feed state IN and receive status updates OUT, but the DB is authoritative.

If GitHub is down, clawdio should still:
- Show all work items and their current state
- Dispatch agents and run workflows
- Track session output and workflow progress
- Queue status sync for when connectivity returns

## What changes

**Work item lifecycle is DB-driven:**
- States (inbox, ready, in_progress, needs_review, done) are owned by the DB
- Transitions happen in the DB first, then get synced to sources
- External state changes (issue closed on GitHub) are sync events that propose DB updates, not direct mutations

**Source sync becomes bidirectional and resilient:**
- Inbound: source proposes state changes (issue closed -> suggest moving to done)
- Outbound: DB state changes get pushed to sources (moved to in_progress -> add label)
- Failed syncs are queued and retried, not dropped
- Sync queue persisted in SQLite so it survives restarts

**PR/review state tracked locally:**
- pr_state (draft, ready, merged, closed) stored on the work item
- ci_status (pending, passing, failing) stored on the work item
- review_verdict (approve, changes_requested) stored on the work item
- Updated by sync, used by the workflow engine without querying GitHub

## Schema additions

```sql
ALTER TABLE work_items ADD COLUMN pr_state TEXT DEFAULT '';
ALTER TABLE work_items ADD COLUMN ci_status TEXT DEFAULT '';
ALTER TABLE work_items ADD COLUMN review_verdict TEXT DEFAULT '';

CREATE TABLE sync_queue (
    id TEXT PRIMARY KEY,
    work_item_id TEXT NOT NULL,
    target TEXT NOT NULL,         -- github, jira, etc.
    action TEXT NOT NULL,         -- add_label, post_comment, transition
    payload TEXT NOT NULL,        -- JSON
    status TEXT DEFAULT 'pending', -- pending, synced, failed
    retry_count INTEGER DEFAULT 0,
    created_at DATETIME DEFAULT CURRENT_TIMESTAMP,
    synced_at DATETIME
);
```

## Impact on other issues

- ~~#86 (Source/Workspace/Target)~~ -- merged, Source.SyncStatus becomes async, writes to sync_queue
- #109 (reopen -> inbox) -- trivial once state is DB-driven
- ~~#101 (parser bug)~~ -- fixed, verdict stored in DB, not re-queried from GitHub
- #100 (Ship It resume) -- all state needed to resume is in the DB

Closes jasonmadigan/clawdio#110


jasonmadigan:
[clawdio review]

Good, I've reviewed the full diff and all new files. Here's the review:

---

## CI Status

**Pending** -- `build-and-test` is still in progress. Cannot confirm pass/fail yet.

## Summary

Adds a persistent sync queue backed by SQLite so that outbound actions to external sources (GitHub labels, comments) are queued and retried rather than fired directly. Also adds a `pr_state` column to the work items table. The `GitHubSource` and `Poller` now route through an `Enqueuer` when one is set, falling back to direct calls when it isn't -- preserving backwards compatibility.

## Findings

### Important

1. **`ensureLabelsOnce` now takes a `workItemID` that doesn't belong there** (`github_source.go:240`). The label-ensure operation is per-repo, not per-item -- it creates the label definitions on the repository. The `workItemID` is passed solely because the enqueuer API requires one. This works but is semantically wrong: the sync queue entry references an arbitrary work item rather than the repo it actually belongs to. If that work item is deleted, the queue entry's `work_item_id` becomes a dangling reference. Not a bug today, but a design smell worth noting.

2. **No backoff between retries** (`processor.go:109-119`). The drain loop processes all pending entries (including failed ones eligible for retry) on every tick. With a 10-second interval and no exponential backoff delay, a persistently failing action will be retried every 10 seconds until it hits `maxRetries=5`. The issue description mentions "exponential backoff" but the implementation is pure retry counting without any delay scaling. For a label-add failure against a down GitHub, this means 5 rapid retries over ~50 seconds then permanent failure -- not particularly resilient.

3. **`json.Marshal` error silently swallowed** (`enqueue.go:58`). `data, _ := json.Marshal(payload)` -- if marshal fails (unlikely with these simple structs, but still), a nil/empty payload gets written to the DB. The entry will then fail on unmarshal in the processor and count against retries. Worth at least logging the error.

### Minor

4. **`sync_queue` table has no index on `status`** (`migrations.go`). The `PendingSyncEntries` query filters by `status` on every drain tick. Fine at small scale, but a `CREATE INDEX` on `(status, retry_count)` would be cheap insurance.

5. **`PostComment` in the public `PostComment` method uses the enqueuer but the private `postComment` helper also checks the enqueuer** (`github_source.go:97-101` vs `github_source.go:317-323`). The public `PostComment` method has its own enqueuer check that short-circuits before reaching the helper. Not a bug -- both paths produce the same result -- but the public method's inline check is now redundant since `postComment` handles it.

6. **Issue says `ci_status` and `review_verdict` columns** but the schema already had `pr_ci_status` and `pr_review_status` from prior work. Only `pr_state` is actually new. The issue description is slightly out of date vs the implementation, but the implementation is correct -- it adds only what's missing.

## Suggestions

- Consider adding backoff logic: skip failed entries whose `created_at + backoff(retry_count)` is in the future. Simple multiplication (`retry_count * 30s`) would prevent hammering a down service.
- Add an index on `sync_queue(status)` in the migration.
- Remove the redundant enqueuer check in the public `PostComment` method since the private `postComment` helper already handles it.

## Verdict

Solid, well-structured change. The sync queue design is clean, the fallback to direct calls when no enqueuer is set preserves backwards compatibility, and the test coverage hits the key paths (enqueue, mark synced, retry counting, purge, PR state). The findings above are all "important but not blocking" -- the backoff gap is the most notable since it undermines the stated resilience goal, but it's a follow-up improvement, not a correctness bug.

Approve once CI passes.

VERDICT: approve

PR #130, opened 2026-04-14
Timeline shows post-action events (push, PR created, CI, merge)

### LINKED ISSUE #123: Timeline should show post-action events (push, PR created, CI, merge), not just sessions
## Problem

The timeline only shows agent session events (dispatch + completion). Post-actions like push, create PR, wait for CI, and merge are invisible. A successfully merged item's timeline ends at "review-pr completed" with no indication that the PR was pushed, reviewed on GitHub, CI passed, or merged.

## What's missing

The full lifecycle for a Ship It run:

```
dispatched: implement-issue     <-- visible
session completed               <-- visible  
pushed changes                  <-- INVISIBLE
PR #99 created                  <-- INVISIBLE
dispatched: review-pr           <-- visible
session completed (approve)     <-- visible
CI checks passed                <-- INVISIBLE
PR marked ready                 <-- INVISIBLE
PR merged                       <-- INVISIBLE
```

## Fix

Store workflow events (post-action executions) as timeline entries. Options:

**Option A: workflow_events table**
```sql
CREATE TABLE workflow_events (
    id TEXT PRIMARY KEY,
    work_item_id TEXT NOT NULL,
    workflow_run_id TEXT,
    event_type TEXT NOT NULL,  -- push, pr_created, ci_passed, merged, etc.
    detail TEXT DEFAULT '',    -- PR URL, branch name, etc.
    created_at DATETIME DEFAULT CURRENT_TIMESTAMP
);
```

The timeline component reads both sessions and workflow_events, merges by timestamp.

**Option B: extend sessions table**
Add a `type` column to sessions: `skill` (current) or `action`. Post-actions create lightweight session records with the action name as skill_name and the result as summary. The timeline already reads sessions.

**Option C: action contracts (#121)**
When actions become contracts, they'll naturally produce records that the timeline can display. This is the cleanest long-term solution but depends on #121.

## Recommendation

Option A for now (simple, doesn't change session semantics). Migrate to Option C when #121 lands.

## Relates to

- #121 (action contracts) -- long-term fix
- #79 (timeline UX, merged) -- timeline already has the rendering, just needs data
- #83 (logging, merged) -- workflow events are logged, just not stored

jasonmadigan:
Closes #123


jasonmadigan (commented):
Good. I have a complete picture now.

---

## PR Review: Timeline shows post-action events (#130)

**CI status:** Passing.

**Summary:** Adds a `workflow_events` table to record post-action lifecycle events (push, PR created, CI pass/fail, merge) that were previously invisible in the timeline. Events are recorded from both the workflow engine (automated flows) and the server handlers (manual actions). The frontend fetches and merges these into the existing timeline, with fallback to the old state-derived entries when no real events exist.

### Findings

**Important:**

1. **Missing index on `workflow_events.work_item_id`** — `ListWorkflowEvents` queries by `work_item_id` on every timeline render. Without an index this is a full table scan. Should add `CREATE INDEX IF NOT EXISTS idx_workflow_events_item ON workflow_events(work_item_id)` in both the migration list and the full schema string.

2. **`currentRunID` scans all runs on every event** — `engine.go:427` calls `ListWorkflowRuns` (which returns all runs for the item) and then iterates to find a running/paused one. This is called multiple times per workflow step execution. Not a problem at current scale, but it would be cleaner to pass the `runID` down from `executeSteps` which already has it, rather than re-querying each time.

3. **No error handling on the `useEffect` fetch** — `Timeline.tsx:154` does `.then()` with no `.catch()`. A failed fetch silently swallows the error. Should add a catch (even just logging to console) to avoid confusion when debugging.

**Minor:**

4. **Duplicate events possible from manual handlers** — If a user triggers push-and-PR via the server handler, both `pushed` and `pr_created` events are recorded. If the same work item later goes through a workflow that also pushes, a second `pushed` event gets recorded. This is probably fine (the timeline shows the actual history of what happened), but worth noting as intentional.

5. **`handleApprovePR` records `merged` without the PR URL** — At `work_item_handlers.go:533`, the merge event is recorded with an empty detail string, while the workflow engine version at `engine.go:400` passes `item.PRURL`. Inconsistent — the handler should pass the PR URL for consistency.

### Suggestions

- Add the `work_item_id` index. This will matter as the events table grows.
- Consider threading the `runID` through the execute methods rather than querying for it. Would simplify the code and remove the N+1 query pattern.
- Add `.catch()` to the `useEffect` fetch.

### Verdict

Clean implementation that matches Option A from the issue exactly. The fallback logic for items that predate the feature is well thought out. Tests cover the key paths. The missing index is the only thing I'd want fixed before merge.

VERDICT: changes_requested

jasonmadigan (commented):
I've seen everything I need. Let me compile the review.

---

## PR Review: Timeline shows post-action events (#130)

**CI status:** Unable to check (GitHub API rate limited). Proceeding with code review.

**Summary:** Adds a `workflow_events` table to record post-action lifecycle events (push, PR created, CI pass/fail, merge) into the timeline. Events are recorded from both the workflow engine (automated flows) and the server handlers (manual actions). The frontend fetches and merges these into the existing timeline, with fallback to state-derived entries for items predating this feature.

### Previous review feedback

The prior review raised 5 findings. Commit `dab36f0` addresses them:

| # | Finding | Status |
|-|-|-|
| 1 | Missing index on `workflow_events.work_item_id` | Fixed — index added in both migration list and full schema |
| 2 | `currentRunID` re-queries all runs per event | Fixed — `runID` threaded through `ExecutePostAction` and all execute methods |
| 3 | No `.catch()` on `useEffect` fetch | Fixed — `.catch(err => console.warn(...))` added |
| 4 | Duplicate events from manual+workflow | Intentionally left as-is (documented in commit message) |
| 5 | `handleApprovePR` missing PR URL on merge event | Fixed — `item.PRURL` now passed in both `handleApprovePR` and `handleMergePR` |

All substantive feedback addressed.

### New findings

**Minor:**

1. **Ignored error on `GetWorkItem` in handlers** — `handleApprovePR` (line 523) and `handleMergePR` (line 554) both call `s.store.GetWorkItem(id)` and discard the error with `item, _ :=`. If the work item somehow doesn't exist (race condition, already deleted), `item` will be zero-valued and `item.PRURL` will be empty. This is low risk since `reviewContext` already validated the item exists a few lines above, but discarding errors silently is a pattern worth cleaning up. Not a blocker.

2. **Extra blank line** — `engine.go:422` has a double blank line between `recordEvent` and `setPhase`. Trivial.

### Verdict

Previous review feedback addressed cleanly. The implementation is solid — well-structured separation between engine events and handler events, clean fallback logic for pre-existing items, and good test coverage. The remaining findings are cosmetic.

VERDICT: approve

PR #106, opened 2026-04-13
Improve logging: structured output, workflow lifecycle, error context

jasonmadigan:
Logging is currently a mix of ad-hoc Printf calls with inconsistent formatting. Hard to follow what's happening during autopilot workflows, session lifecycle, and error scenarios.

## Problems

- **No structure** -- plain text lines with inconsistent prefixes (sometimes "autopilot:", sometimes "workflow:", sometimes "session %s:")
- **Workflow lifecycle is hard to follow** -- individual events log but there's no summary of "item X: implement -> push -> PR #80 -> review -> approve -> merge" as a cohesive flow
- **Errors lack context** -- "CI check: context deadline exceeded" doesn't say which item, which PR, which workflow run
- **No log levels** -- everything is Printf. Can't filter debug noise from important events
- **Session events are noisy** -- "event type=assistant" on every stream-json line floods the log. Useful for debugging, useless in normal operation
- **Poller logs are repetitive** -- "skipping cleanup of X (active session)" every 30 seconds
- **No request ID tracing** -- can't correlate an API request through dispatch -> session -> post-action -> workflow

## What good looks like

**Structured log lines** with consistent fields:
\`\`\`
level=info component=autopilot item=clawdio#62 action=dispatched skill=implement-issue
level=info component=workflow run=95b5975c step=2/3 action=review-pr status=completed verdict=approve
level=info component=workflow run=95b5975c step=3/3 action=merge pr=#80 status=success
level=error component=workflow run=e1b01684 step=3/3 action=merge error="CI checks failed"
level=debug component=session id=c952ec55 event=assistant
\`\`\`

**Workflow summary on completion:**
\`\`\`
level=info component=autopilot item=clawdio#62 workflow=implement-and-land result=merged pr=#80 duration=3m42s
\`\`\`

## Scope

- **Session manager** -- reduce event-level noise to debug, keep dispatch/complete/error at info
- **Autopilot engine** -- log rule matches, dispatch decisions, capacity state
- **Workflow engine** -- log step transitions, post-actions, review verdicts, final outcome with duration
- **Poller** -- reduce repetitive "skipping cleanup" to debug, keep import/sync at info
- **Server** -- request logging with method, path, duration, status code
- **Health check** -- log when stuck sessions are detected and cleaned up

## Implementation

Could use \`log/slog\` (stdlib since Go 1.21) for structured logging. Or keep Printf but with a consistent format. slog is cleaner but a bigger change.

[...]

PR #105, opened 2026-04-13
Autopilot prioritisation: triage skill to order dispatch queue

jasonmadigan:
Autopilot currently dispatches matching inbox items in arbitrary order (whatever the DB returns). It should use a configurable triage step to prioritise which items to work on first.

## How it works

When autopilot has multiple matching items and limited slots (max_concurrent), it needs to decide which items to dispatch first. A triage skill runs a quick classification pass over the pending items and returns a ranked order.

## Approach

1. **Triage skill** -- a dispatch skill (e.g. \`skills/triage.md\`) that receives a batch of items and returns them ranked by priority. Could use heuristics (labels, age, repo, type) or a quick Claude call for smarter classification.

2. **Integration with autopilot** -- before dispatching, autopilot passes the matched items through the triage skill. The skill returns an ordered list. Autopilot dispatches from the top.

3. **Configurable** -- the triage skill name is set in the autopilot config:
\`\`\`json
"autopilot": {
  "triage_skill": "triage",
  ...
}
\`\`\`
If not set, items dispatch in arrival order (current behaviour).

## Triage signals

The skill could consider:
- Labels (bug vs feature vs docs)
- Age (older items first, or newer?)
- Repo priority (some repos matter more)
- Item type (issues vs PR reviews -- reviews might be time-sensitive)
- Size hints (PR diff size, issue description length)
- Dependencies (does this unblock other items?)

## Relates to

- #24 (auto-triage: classify and pre-select skills) -- this is the prioritisation half of that issue
- #23 (autopilot) -- this enhances autopilot's dispatch ordering

Closes jasonmadigan/clawdio#76

PR #104, opened 2026-04-13
Dockerfile and container image for easy onboarding

jasonmadigan:
Create a Docker image with everything needed to run clawdio. Goal: try it in 60 seconds.

## Usage

\`\`\`bash
docker run -it \\
  -e ANTHROPIC_API_KEY=sk-... \\
  -v ~/.config/gh:/home/clawdio/.config/gh:ro \\
  -v clawdio-data:/data \\
  -p 8077:8077 \\
  ghcr.io/jasonmadigan/clawdio:latest
\`\`\`

Open http://localhost:8077, configure repos, start dispatching.

## What's in the image

- clawdio binary (with embedded frontend, depends on #63)
- claude CLI
- gh CLI (authenticated via mounted config)
- tmux (for session management, relates to #58)
- git
- sqlite

## Credentials

- **Anthropic API key:** env var \`ANTHROPIC_API_KEY\`
- **GitHub auth:** mount \`~/.config/gh\` read-only. The gh CLI picks it up automatically
- **Claude Code config:** optionally mount \`~/.claude\` for skills, CLAUDE.md, etc.

## Persistence

- \`/data\` volume for SQLite DB, worktrees, and config
- Named volume survives container restarts
- Config file at \`/data/config.json\` (or generated from env vars on first run)

## Image build

Multi-stage Dockerfile:
1. **ui-build stage:** Node, \`npm ci && npm run build\`
2. **go-build stage:** Go, copy ui/dist, \`go build\`
3. **runtime stage:** minimal base (debian-slim or alpine), install claude CLI, gh CLI, tmux, git, copy binary

## Future

- CI to build and push to ghcr.io on tag/release
- docker-compose.yml for local dev with volume mounts
- ARM64 support for running on cloud instances

## Depends on

- #63 (embedded frontend) -- single binary, no Node at runtime
- #58 (tmux wrapping) -- nice to have for peek/persistence inside the container

Closes jasonmadigan/clawdio#64

PR #103, opened 2026-04-13
Production build: embed frontend in Go binary, single port

jasonmadigan:
Currently dev runs two servers: Go on :8077 and Vite on :5173 with proxy. For production/Docker, the Go server should serve the built frontend static files directly.

## What changes

- `make build` runs `cd ui && npm run build` first, then builds the Go binary
- Go binary embeds \`ui/dist/\` using \`embed.FS\`
- Server serves static files from the embedded FS at \`/\`, API at \`/api\`, WebSocket at \`/ws\`
- Single binary, single port, no Node runtime needed at runtime

## Implementation

Go 1.16+ \`//go:embed\` directive. The server already has the API routes registered -- just add a catch-all file server for the frontend SPA (serve index.html for any non-API route).

Build sequence in Makefile:
\`\`\`
build:
	cd ui && npm run build
	go build -o clawdio ./cmd/clawdio
\`\`\`

## Why

- Docker image doesn't need Node
- Single port simplifies networking and reverse proxy setup
- One binary to distribute
- Dev workflow unchanged (still use make dev-api + make dev-ui for HMR)

Closes jasonmadigan/clawdio#63

PR #134, opened 2026-04-14
Workflow: add rebase-and-resolve-conflicts step before merge

### LINKED ISSUE #107: Workflow: add rebase-and-resolve-conflicts step before merge
When multiple Ship It workflows run in parallel, the first to merge makes the rest conflict. The workflow should automatically rebase and resolve conflicts before attempting merge.

Add a rebase action to the implement-and-land workflow, between review-approved and merge:

\`\`\`yaml
check_ci:
  action: target.wait_for_ci
  on_pass: rebase
rebase:
  action: workspace.rebase_on_main
  on_success: merge
  on_conflict: resolve_conflicts
resolve_conflicts:
  skill: resolve-conflicts
  on_complete: rebase
\`\`\`

The resolve-conflicts skill is an AI agent that reads the conflict markers and resolves them, same as a human would. If it can't resolve, the workflow pauses.

Observed today: PRs #104 and #106 both had CI green and approve verdicts but couldn't merge because #103 and #105 merged first, creating conflicts.

jasonmadigan:
When multiple Ship It workflows run in parallel, the first to merge makes the rest conflict. The workflow should automatically rebase and resolve conflicts before attempting merge.

Add a rebase action to the implement-and-land workflow, between review-approved and merge:

\`\`\`yaml
check_ci:
  action: target.wait_for_ci
  on_pass: rebase
rebase:
  action: workspace.rebase_on_main
  on_success: merge
  on_conflict: resolve_conflicts
resolve_conflicts:
  skill: resolve-conflicts
  on_complete: rebase
\`\`\`

The resolve-conflicts skill is an AI agent that reads the conflict markers and resolves them, same as a human would. If it can't resolve, the workflow pauses.

Observed today: PRs #104 and #106 both had CI green and approve verdicts but couldn't merge because #103 and #105 merged first, creating conflicts.

Closes jasonmadigan/clawdio#107


jasonmadigan:
[clawdio review]

---

## CI status

Passing (build-and-test: 41s).

## Summary

Adds a `rebase_on_main` action between review approval and merge in the implement-and-land workflow. When rebase hits conflicts, it dispatches a `resolve-conflicts` skill (AI agent) to fix them, then retries the rebase. If unresolvable, the workflow pauses. Also refactors `executeAndAdvance` from single-shot to a loop so sequential actions (rebase → merge) chain without recursion.

## Findings

### Important

1. **`executeRebase` returns error on failure, bypassing the `"failed"` transition** (`engine.go:387`). When `git rebase` fails for non-conflict reasons, `executeRebase` returns `("failed", fmt.Errorf(...))`. But `executeAndAdvance` treats any error from `ExecutePostAction` as fatal and returns immediately -- `PostActionDone` never runs, so the `failed → pause` transition in the action contract is dead code. The workflow errors out instead of gracefully pausing.

   This works by accident right now (the workflow still stops), but the intent of the contract is to pause, not crash. Either return `("failed", nil)` like the conflict path does, or handle the error upstream before calling `PostActionDone`. The test `TestPostActionDoneRebaseFailed` tests the engine's transition logic in isolation but doesn't exercise the actual execution path through `executeAndAdvance`.

2. **Hardcoded action name in `PostActionDone`** (`engine.go:128-137`). The `rebase_on_main` action gets a special `if` block to look up its contract transitions. Other contract-based actions would need similar blocks. This is fine for now since it's the only action with outcome-based transitions, but it's worth noting as a pattern that won't scale. Not blocking.

### Minor

3. **`/ fallback:` comment typo** (`engine.go:252`). Missing the second slash -- should be `// fallback:`.

4. **Em dash in comment** (`manager.go:1414`). The comment uses `→` (arrow character). Harmless but inconsistent with the rest of the codebase which uses `->` or plain English.

## Suggestions

- For finding #1: change `executeRebase` to return `("failed", nil)` on rebase failure (non-conflict), matching the conflict path. This lets the contract transition handle it cleanly. The error details are already captured via `recordEvent`.

## Verdict

The architecture is sound -- action contracts with transitions, the chaining loop in `executeAndAdvance`, the resolve-conflicts skill design, and the test coverage are all well thought out. The error-path issue (#1) is the only thing worth addressing.

VERDICT: changes_requested

jasonmadigan:
[clawdio review]

---

## CI status

Passing (build-and-test: 35s).

## Summary

Adds a `rebase_on_main` action between review approval and merge in the implement-and-land workflow. When rebase hits conflicts, a `resolve-conflicts` skill is dispatched to fix them, then the rebase retries. If the conflict can't be resolved, the workflow pauses. Also refactors `executeAndAdvance` from single-shot to a loop so sequential actions (rebase -> merge) chain without recursion.

A prior review raised 4 findings. The follow-up commit (`8b2e978`) addresses #1 (error path) and #4 (unicode arrow). #2 (hardcoded action name) was acknowledged as non-blocking. #3 (comment typo) was not reproducible.

## Findings

### Important

1. **`git fetch` failure still returns an error, bypassing contract transitions** (`engine.go:368`). The review fix correctly changed the rebase failure path to `return "failed", nil` (line 387), but the `git fetch` failure on line 368 still returns `return "failed", fmt.Errorf(...)`. Since `executeAndAdvance` treats any error as fatal, a fetch failure skips `PostActionDone` and the `failed -> pause` transition never fires. This should be `return "failed", nil` for the same reason as the rebase path fix.

### Minor

2. **Resolve-conflicts skill could loop indefinitely**. The cycle is: `rebase_on_main` (conflict) -> `resolve-conflicts` (resolved) -> `rebase_on_main`. If the conflict keeps reappearing (e.g. main keeps moving), there's no iteration limit. Not urgent -- in practice main doesn't move that fast -- but worth a TODO.

3. **No `force-push` after successful conflict resolution**. After `resolve-conflicts` resolves and the rebase succeeds, the rebased branch needs a force-push before merge. The existing `push` action does `git push origin HEAD` without `--force-with-lease`. Unclear if the merge action's `MarkReady`/`gh pr merge` handles this, or if it'll fail because the remote branch diverges from the rebased local. Worth verifying manually.

## Suggestions

- For finding #1: change line 368 to `return "failed", nil` to match the fix at line 387.
- For finding #3: verify the end-to-end path: resolve conflicts -> rebase succeeds -> merge. If the remote branch needs updating, the `rebase_on_main` action contract may need a `push` step after success, or the merge action needs to handle diverged branches.

## Verdict

Finding #1 is a genuine bug in the same class as the one already fixed. The other findings are minor/operational. Good work overall -- the architecture is clean and well tested.

VERDICT: changes_requested

jasonmadigan:
[clawdio review]

The tmux test failure is pre-existing and unrelated to this PR (it requires a running tmux server). The workflow tests all pass.

Here's my review:

---

## CI status

Passing (build-and-test: 37s). The tmux test failure is pre-existing and unrelated.

## Summary

Adds a `rebase_on_main` action between review approval and merge in the implement-and-land workflow. When rebase hits conflicts, it dispatches a `resolve-conflicts` skill (AI agent) to fix them, then retries the rebase. If unresolvable, the workflow pauses. Also refactors `executeAndAdvance` from single-shot to a loop so sequential actions (rebase -> merge) chain without recursion.

Two prior reviews raised findings. The follow-up commits (`8b2e978`, `db4f13c`) addressed the error-path issues (returning `nil` error for contract transitions) and the unicode arrow. Those fixes look correct.

## Findings

### Important

1. **Missing push between rebase and merge.** The rebase happens locally (`git rebase origin/main` in the worktree), but the merge uses `gh pr merge --squash` which operates on the *remote* branch. After a local rebase, the remote branch is stale. The transition `rebased -> merge` chains directly to `executeMerge` without pushing the rebased branch. In the conflict case (the exact scenario this feature targets), `gh pr merge` will hit the same conflict on GitHub's side because it's merging the old remote branch.

   The `Publish` method already handles force-push with `--force-with-lease`, so adding a push step between rebase and merge should be straightforward. Either change the transition to `rebased -> push` (with push chaining to merge), or have `executeRebase` push after a successful rebase before returning `"rebased"`.

### Minor

2. **No iteration limit on the resolve-conflicts loop.** The cycle `rebase_on_main` (conflict) -> `resolve-conflicts` (resolved) -> `rebase_on_main` could loop indefinitely if main keeps moving. Not urgent in practice but worth a TODO.

3. **Hardcoded `"rebase_on_main"` in `PostActionDone`** (`engine.go:128`). The second review noted this too. Fine for now since it's the only action with outcome-based transitions, but this should generalise when more contract-based actions are added.

## Suggestions

- For finding #1: the cleanest fix is probably changing the rebase contract to `rebased: push` and adding a `push -> merge` chain, or having `executeRebase` call `target.Publish` before returning `"rebased"`. Either way the rebased branch needs to reach the remote before `gh pr merge` runs.

## Verdict

The architecture is clean, test coverage is good, and the prior review findings were properly addressed. The missing push between rebase and merge (#1) is a real gap that would cause the feature to fail in the exact scenario it's designed for (conflicting branches). Requesting changes for that.

VERDICT: changes_requested

Issue #131, opened 2026-04-14
Action contracts should support retry for transient failures


jasonmadigan:
Workflow post-actions (push, create_pr, merge) fail permanently on transient errors like GitHub API timeouts or DNS blips. The workflow errors and the item gets stuck.

Observed: push succeeded but create_pr failed with "error connecting to api.github.com". The workflow errored with no retry. The implementation was complete but never got a PR.

Action contracts (#121) should declare retry behaviour:

\`\`\`yaml
name: create_pr
type: action
retry:
  max_attempts: 3
  backoff: 10s
  retryable_errors: [connection, timeout, 5xx]
\`\`\`

The workflow engine retries transient failures automatically, only erroring on persistent failures or after max attempts.

Relates to #121 (action contracts).

Issue #116, opened 2026-04-14
gh CLI authentication in Docker containers


jasonmadigan:
## Problem

The Docker container needs gh CLI authenticated to poll GitHub and create PRs. Current approach: mount `~/.config/gh` from the host. This works if the user is already logged in on their machine, but fails if they're not.

### 1. Mount host credentials (current)

```bash
docker run -v ~/.config/gh:/home/clawdio/.config/gh:ro ...
```

Pros: zero setup if already logged in. No secrets in env vars.
Cons: fails for users who aren't logged in. Read-only mount means the container can't refresh tokens.

### 2. GitHub token via env var

```bash
docker run -e GH_TOKEN=ghp_... ...
```

Pros: simple, works everywhere, no mount needed.
Cons: PAT tokens have broad scope. User has to create and manage the token.

### 3. Auth flow via the UI

When gh isn't authenticated, show a setup step in the UI:
- "GitHub not connected" banner
- Click to start `gh auth login --web` flow
- UI shows the device code and link
- User authenticates in their browser
- Credentials saved inside the container volume

Pros: best UX, no pre-setup needed.
Cons: needs the UI to interact with the gh auth flow. Device code flow needs terminal-like interaction.

### 4. Read-write mount with fallback

Mount `~/.config/gh` read-write. If not authenticated, the container runs `gh auth login` on first start and the creds persist on the host.

```bash
docker run -v ~/.config/gh:/home/clawdio/.config/gh ...
```

Pros: works for both logged-in and not-logged-in users. Creds persist.
Cons: modifying host filesystem from a container feels wrong.

## Recommendation

Support all four, in priority order:
1. `GH_TOKEN` env var (simplest, works in CI/headless)
2. Mount `~/.config/gh:ro` (zero setup for logged-in users)
3. Auth flow in UI (best experience for new users)
4. `docker exec -it <container> gh auth login` as escape hatch

The docker-entrypoint.sh should check for auth on startup and print clear instructions if not authenticated.

## Relates to

- #64 (Dockerfile, merged) -- Docker quickstart guide

PR #139, opened 2026-04-14
Replace aggressive polling with on-demand fetch and optional background sync

### LINKED ISSUE #132: Replace aggressive polling with on-demand fetch and optional background sync
## Problem

The poller runs every 30 seconds across all configured repos, burning through GitHub API rate limits. Hit rate limits today during normal use. For a single-user tool this is wasteful.

## Current behaviour

- Polls every 30s (configurable but default is aggressive)
- Each poll cycle: list issues, list PRs, list notifications per repo per query
- Status sync checks every item on every cycle
- No awareness of rate limits

## Proposed behaviour

**On-demand fetch:**
- Poll once when the app loads (or first page visit)
- Manual refresh button on the board (already exists, just needs to trigger a poll)
- No background polling by default

**Optional background sync:**
- Configurable interval, default off or very slow (5-10 minutes)
- Respects rate limit headers -- back off when approaching limits
- Only fetches repos visible on the current board view

**Webhooks (long-term):**
- GitHub webhooks push events to clawdio instead of clawdio pulling
- Eliminates polling entirely for GitHub
- Needs a publicly reachable endpoint (or smee.io/ngrok for local dev)
- Best fit for the Docker/container deployment story

**Rate limit awareness:**
- Check \`X-RateLimit-Remaining\` header on every \`gh\` response
- If below threshold, pause polling until reset
- Show rate limit status in the UI (settings page or status bar)
- Log warnings when approaching limits

## Relates to

- #110 (DB as source of truth) -- local state reduces need for constant polling
- #131 (retry for transient failures) -- rate limit is a retryable condition
- #116 (gh auth in Docker) -- webhooks need auth setup

jasonmadigan:
## Problem

The poller runs every 30 seconds across all configured repos, burning through GitHub API rate limits. Hit rate limits today during normal use. For a single-user tool this is wasteful.

## Current behaviour

- Polls every 30s (configurable but default is aggressive)
- Each poll cycle: list issues, list PRs, list notifications per repo per query
- Status sync checks every item on every cycle
- No awareness of rate limits

## Proposed behaviour

**On-demand fetch:**
- Poll once when the app loads (or first page visit)
- Manual refresh button on the board (already exists, just needs to trigger a poll)
- No background polling by default

**Optional background sync:**
- Configurable interval, default off or very slow (5-10 minutes)
- Respects rate limit headers -- back off when approaching limits
- Only fetches repos visible on the current board view

**Webhooks (long-term):**
- GitHub webhooks push events to clawdio instead of clawdio pulling
- Eliminates polling entirely for GitHub
- Needs a publicly reachable endpoint (or smee.io/ngrok for local dev)
- Best fit for the Docker/container deployment story

**Rate limit awareness:**
- Check \`X-RateLimit-Remaining\` header on every \`gh\` response
- If below threshold, pause polling until reset
- Show rate limit status in the UI (settings page or status bar)
- Log warnings when approaching limits

## Relates to

- #110 (DB as source of truth) -- local state reduces need for constant polling
- #131 (retry for transient failures) -- rate limit is a retryable condition
- #116 (gh auth in Docker) -- webhooks need auth setup

Closes jasonmadigan/clawdio#132


jasonmadigan:
[clawdio review]

Good. I've read the full diff and key files. Here's my review:

---

## PR Review: #139 — Replace aggressive polling with on-demand fetch and optional background sync

**CI status:** Pending (build-and-test still running). No existing reviews or comments.

**Summary:** Replaces the 30s polling loop with an on-demand model: poll once at startup, then only when the user clicks refresh. Background polling is opt-in via `background_poll` config. Adds rate limit awareness (`gh api rate_limit`) and surfaces the remaining quota in the UI header.

### Findings

**Important:**

1. **`r.Context()` passed to background goroutine** — `poll_handlers.go:16` does `go s.pollCoord.PollOnce(r.Context())`. The request context is cancelled when the HTTP response is sent (which happens immediately on the next line). The goroutine inherits a context that's already dead or about to be. This should use `context.Background()` or a server-scoped context instead.

2. **Double rate limit check per poll** — `PollOnce` calls `isRateLimited()` before polling (which calls `CheckRateLimit`, i.e. `gh api rate_limit`) and then `checkAndStoreRateLimit()` after. That's two extra `gh` CLI invocations per poll cycle just for bookkeeping. The pre-check is valuable, but the post-check could read headers from the poll response instead, or be dropped entirely since the pre-check on the *next* poll will catch the updated state. Not a bug, but unnecessary API usage for a PR that's about reducing API usage.

3. **`StartBackground` return value is ignored** — `main.go:281` calls `coord.StartBackground(ctx)` but discards the returned stop function. The background goroutine will leak on shutdown. Either capture and defer the stop, or wire it into the server's shutdown path.

**Minor:**

4. **`paused` field is written but never read** — `coordinator.go:35` declares a `paused` field that gets set in `isRateLimited` and `checkAndStoreRateLimit`, but nothing ever checks it. Dead state. Either expose it (e.g. in the rate limit API response) or remove it.

5. **Frontend poll delay is arbitrary** — `App.tsx:49` uses `setTimeout(() => { ... }, 1000)` to wait for the async poll. This is fragile — on slow connections or large repos the poll won't have finished. A WebSocket broadcast already exists for item updates, so the `loadItems()` call after the timeout may be redundant. Consider either trusting the WebSocket broadcast or returning a poll ID the frontend can track.

6. **Config default differs between code and docs** — `docs/configuration.md` says poll_interval default is `5m`, and `coordinator.go:123` hardcodes `5 * time.Minute` as fallback. The old default in the removed code was `2 * time.Minute`. The new default is fine, just noting the intentional change from 2m to 5m is only visible in the diff, not called out in the PR description.

### Suggestions

- For item 1, change to `go s.pollCoord.PollOnce(context.Background())` or pass in a server-level context.
- For item 2, drop `checkAndStoreRateLimit` — the pre-poll check in `isRateLimited` already stores the rate limit, and the next poll's pre-check will refresh it.
- For item 3, capture the stop function: `stop := coord.StartBackground(ctx); defer stop()` or similar.
- For item 4, either expose `paused` in the `/api/rate-limit` response or delete the field.

### Verdict

Clean decomposition. The Coordinator abstraction is sound and the concurrency guard in PollOnce is correct. The context bug in the poll handler is the main thing to fix before merging.

VERDICT: changes_requested

jasonmadigan:
[clawdio review]

Good, I've read the full diff and key files. Here's my review:

---

## PR Review: #139 — Replace aggressive polling with on-demand fetch and optional background sync

**CI status:** Pending (build-and-test still running). No existing reviews or comments.

**Summary:** Replaces the 30s polling loop with an on-demand model: poll once at startup, then only when the user clicks refresh. Background polling is opt-in via `background_poll` config. Adds rate limit awareness (`gh api rate_limit`) and surfaces the remaining quota in the UI header.

### Findings

**Important:**

1. **`cancel` field is stored but never used for cleanup** — `coordinator.go:33` stores `cancel` on the struct, but nothing reads it outside `StartBackground`. The stop function returned by `StartBackground` already captures `cancel` in its closure, so the struct field is dead state. It's harmless but adds confusion — either use it (e.g. for a `Stop()` method) or drop it from the struct.

2. **Rate limit check adds a `gh api` call before every poll** — `isRateLimited` in `coordinator.go:164` shells out to `gh api rate_limit` before each poll. For on-demand use (user clicks refresh) this means two `gh` invocations per refresh: one for rate limit, one for the actual poll. Since the rate limit info is refreshed in the response headers of any GitHub API call, consider caching the rate limit from the previous poll's response instead, and only doing the explicit check when `lastRate` is nil or stale (e.g. past reset time). Not a bug, but counterproductive for a PR about reducing API usage.

3. **`paused` field referenced in the review template's earlier analysis doesn't exist** — I can confirm the actual code doesn't have a `paused` field. The earlier automated review comment was wrong about this. The struct is clean.

**Minor:**

4. **Frontend poll-then-loadRateLimit timing** — `App.tsx:51-54`: `triggerPoll` fires `POST /api/poll` (async, returns immediately) then calls `loadRateLimit()`. Since the poll hasn't finished yet, the rate limit returned will be the pre-poll value. This is fine functionally (the badge updates on next load or next refresh), but the comment "items arrive via WebSocket broadcast; just refresh rate limit" implies the rate limit will reflect the poll that just happened, which it won't. Consider refreshing rate limit on the WebSocket broadcast event instead, or just removing the `loadRateLimit()` call since it's showing stale data anyway.

5. **`New` constructor changed to accept `checker` instead of `source`** — `coordinator_test.go:33` passes a `mockChecker` to `poller.New`, but the coordinator passes `ghSource` as the checker argument in `main.go:267`. This works because `GitHubSource` satisfies the checker interface, but the test and production code use different types, making the test less representative. Not blocking, just noting.

6. **Config default change not called out** — The old default poll interval was `2 * time.Minute` (hardcoded in the removed `startPoller`). The new default is `5 * time.Minute` in `coordinator.go:119`. This is sensible but worth mentioning in the PR description since it changes existing behaviour for anyone who had the old default.

### Suggestions

- For finding 2: add a staleness check — skip the `gh api rate_limit` call if `lastRate` is non-nil and `lastRate.ResetAt` is in the future. Only shell out when there's no cached value or the reset time has passed.
- For finding 4: either refresh rate limit from the WebSocket handler, or accept the badge shows slightly stale data and drop the explicit `loadRateLimit()` after `triggerPoll`.

### Verdict

Clean decomposition overall. The Coordinator abstraction is well-structured, concurrency guard in `PollOnce` is correct, and the wiring in `main.go` properly captures the stop function. The rate limit pre-check on every poll is the main thing worth addressing — it's an extra API call that partially undermines the goal of reducing API usage.

VERDICT: changes_requested

jasonmadigan:
[clawdio review]

I've now thoroughly reviewed all the code. Here's my review:

---

## PR Review: #139 -- Replace aggressive polling with on-demand fetch and optional background sync

**CI status:** Pending (build-and-test still running).

**Summary:** Replaces the 30s polling loop with an on-demand model: poll once at startup, then only when the user clicks refresh. Background polling is opt-in via `background_poll` config. Adds rate limit awareness with caching, and surfaces remaining quota in the UI header.

### Prior review findings -- status

Two automated reviews raised issues. All have been addressed in the follow-up commits:
- Context bug (`r.Context()` in goroutine) -- fixed, uses `context.Background()` with clear comment
- Double rate limit API call -- fixed, `isRateLimited` now uses cached values within the reset window
- Stop function leak -- fixed, `main.go:230` captures and defers `stopPoll`
- Dead `paused`/`cancel` fields -- removed
- Frontend timing issue -- fixed, rate limit now refreshes via WebSocket broadcast handler, no more arbitrary `setTimeout`

### New findings

**Minor:**

1. **Rate limit cache never updates after initial check** -- `isRateLimited` stores the rate limit on the first call, then uses the cache until `ResetAt` passes (typically 1 hour). But the actual `Remaining` count decreases with every poll. The cached value becomes increasingly stale -- if you poll 20 times in an hour, the cache still shows the `Remaining` from the first check. This is acceptable for the threshold check (100 is a generous buffer), but the UI badge at `GET /api/rate-limit` will show a stale value too. Not blocking, but worth knowing. A future improvement could decrement `lastRate.Remaining` after each successful poll.

2. **`handleRateLimit` falls through to live fetch** -- `poll_handlers.go:29-36`: when `pollCoord.RateLimit()` returns nil (no poll has run yet), it calls `ghpkg.CheckRateLimit(r.Context())` inline in the request handler. This is fine, but means the first load of the settings page or any page that calls `/api/rate-limit` before the startup poll completes will block on a `gh api` call. Shouldn't be noticeable in practice since the startup poll usually finishes fast, but the request could hang if `gh` is slow.

3. **`config.example.json` has `background_poll: false`** -- Good default, but it means anyone upgrading from before this PR will get background polling silently disabled (their existing config won't have the field, and Go's zero value for bool is `false`). This is actually the desired behaviour per the issue, but worth mentioning in release notes.

### Verdict

Clean work. The Coordinator abstraction is well-structured, concurrency is handled correctly, the rate limit caching addresses the earlier review feedback properly, and the frontend wiring through WebSocket broadcast is the right approach. The remaining observations are minor and don't block merge. Waiting on CI to confirm.

VERDICT: approve

Issue #114, opened 2026-04-14
Issue-centric UX: PR actions and status on the parent issue panel, not separate work items


jasonmadigan:
## Problem

When working on an issue, the workflow creates a PR as a separate work item. To action the PR (review, merge, see CI status), you need to click away from the issue to the PR work item. This is a jarring context switch -- the issue is what you care about, the PR is just an artefact of working on it.

## Desired behaviour

The issue panel is the home for the full lifecycle:

- Issue description and timeline
- Linked PR section showing: PR title, status (draft/open/merged), CI status, review verdict
- PR action buttons directly on the issue panel: push, create PR, review, approve, merge
- Agent output from all sessions (implement, review, address-feedback) in one timeline
- No need to switch to the PR work item to do anything

## When PR work items still appear standalone

PR work items should only show as standalone board cards when they're orphans:
- An external PR assigned to you for review (no parent issue in your board)
- A PR from another repo with no linked issue

When a PR has a parent issue, it renders as a nested element under the parent (already done via #52 grouped cards), and all actions live on the parent's panel.

## What changes

**Panel.tsx:**
- When the selected item has a linked PR (via childrenMap or pr_url), show PR status and actions inline
- PR actions (push, review, merge) work on the pr_url, not the current item's source_ref
- Timeline shows sessions from both the issue and its linked PR work items

**Board.tsx:**
- PR work items with a parent_id are already hidden from the top-level board (#52)
- No change needed here

**Workflow engine:**
- Already works on the issue work item, not the PR work item
- pr_url on the issue drives the review/merge steps

## Relates to

- #52 (grouped board cards, merged) -- visual nesting is done, but actions are still on the PR panel
- #113 (multiple PRs per issue, closed) -- the issue panel shows the active PR
- #110 (DB source of truth) -- PR state tracked on the issue work item

PR #142, opened 2026-04-14
UI controls for work item state: pause, resume, reset, cancel autopilot

### LINKED ISSUE #136: UI controls for work item state: pause, resume, reset, cancel autopilot
## Problem

There's no way to manage work item state from the UI beyond what the post-action buttons offer. When things go wrong (stuck workflows, failed autopilot, wrong state), the only fix is manual DB edits.

## Needed controls

**State transitions:**
- Move item between columns manually (drag-and-drop exists but explicit buttons too)
- Reset to inbox (clear all workflow/autopilot state, start fresh)
- Mark as done (skip the workflow, just close it)

**Autopilot controls:**
- Pause autopilot on a running item (stop after current step, don't advance)
- Resume paused autopilot
- Cancel autopilot (clear mode, keep current state)
- Retry failed step (re-dispatch the skill that errored)

**Workflow controls:**
- Show current workflow state (which step, what phase)
- Skip current step (advance to next)
- Restart workflow from a specific step
- Cancel workflow entirely

These should be accessible from the panel, not buried in settings. A dropdown or context menu on the work item card would work.

## Relates to

- #119 (stuck workflow recovery) -- manual recovery is needed until auto-recovery works
- #100 (Ship It resume) -- resuming from a specific state
- #114 (issue-centric UX) -- all controls on the issue panel

jasonmadigan:
## Problem

There's no way to manage work item state from the UI beyond what the post-action buttons offer. When things go wrong (stuck workflows, failed autopilot, wrong state), the only fix is manual DB edits.

## Needed controls

**State transitions:**
- Move item between columns manually (drag-and-drop exists but explicit buttons too)
- Reset to inbox (clear all workflow/autopilot state, start fresh)
- Mark as done (skip the workflow, just close it)

**Autopilot controls:**
- Pause autopilot on a running item (stop after current step, don't advance)
- Resume paused autopilot
- Cancel autopilot (clear mode, keep current state)
- Retry failed step (re-dispatch the skill that errored)

**Workflow controls:**
- Show current workflow state (which step, what phase)
- Skip current step (advance to next)
- Restart workflow from a specific step
- Cancel workflow entirely

These should be accessible from the panel, not buried in settings. A dropdown or context menu on the work item card would work.

## Relates to

- #119 (stuck workflow recovery) -- manual recovery is needed until auto-recovery works
- #100 (Ship It resume) -- resuming from a specific state
- #114 (issue-centric UX) -- all controls on the issue panel

Closes jasonmadigan/clawdio#136


jasonmadigan:
GitHub API is temporarily unreachable, but CI was pending when I checked earlier. I have everything I need for the review.

---

## Review: PR #142 -- UI controls for work item state management

### CI Status

**Pending** -- the `build-and-test` check was still running when last checked. GitHub API is currently unreachable for a re-check.

### Summary

Adds a kebab menu (three-dot) to the panel header with "Reset to inbox", "Mark as done", and "Move to ready" actions. Also adds "Cancel workflow", "Retry step", and "Dismiss" buttons inline on the workflow progress bar. Four new backend endpoints support these operations. The frontend shows errored workflow runs (not just running/paused) so retry/dismiss controls are visible.

### Findings

**Important:**

1. **`handleResetItem` updates item fields then calls `UpdateWorkItem`, but also calls `UpdateWorkItemAutopilotMode` separately beforehand.** The `item.AutopilotMode = false` is set on the struct before `UpdateWorkItem`, but `UpdateWorkItemAutopilotMode` was already called. This double-write is harmless but redundant -- pick one path. The `UpdateWorkItemAutopilotMode` call can be removed since `UpdateWorkItem` will persist the struct with `AutopilotMode = false` anyway. (`work_item_handlers.go:786-793`)

2. **`handleRetryStep` sets item status to `StatusInProgress` before dispatching, but on dispatch failure rolls back to `StatusReady`.** If the item was in a different status before (e.g. `inbox` or `done`), the rollback to `StatusReady` is wrong -- it should restore the original status. (`work_item_handlers.go:923-924`)

3. **Errored workflow runs are now surfaced in the panel, but the display logic picks the first errored run if no active run exists.** If there are multiple errored runs (e.g. from previous retries), the most recent one wins because `ListWorkflowRuns` is `ORDER BY created_at DESC`. That's correct, but worth a brief comment since the fallback chain (`running/paused || errored`) is now non-obvious. (`Panel.tsx:367-370`)

**Minor:**

4. **The `ItemActionsMenu` doesn't have a "Cancel autopilot" option.** The issue specifically requests pause/resume/cancel autopilot controls in the dropdown. Currently autopilot cancel only exists via the existing `DELETE /api/work-items/{id}/autopilot` endpoint but there's no menu entry for it. This is a gap vs the issue spec.

5. **No confirmation on destructive actions.** "Reset to inbox" cancels active sessions and wipes workflow state. "Mark as done" also cancels sessions. These are one-click with no undo. A confirmation step (even `window.confirm`) would be reasonable for reset at minimum.

6. **"Move to ready" uses the generic `updateItem` PATCH endpoint** while the other actions use dedicated endpoints. This means it doesn't cancel active sessions or workflows -- moving an in-progress item to "ready" could leave orphaned sessions running. Either add a dedicated endpoint that cleans up, or hide this option when sessions are active.

7. **Missing `Escape` key handler on the dropdown.** The menu closes on outside click but not on Escape, which is the standard keyboard dismissal pattern.

### Suggestions

- Remove the redundant `UpdateWorkItemAutopilotMode` call in `handleResetItem` (finding 1).
- Save the original item status before the retry dispatch attempt so rollback restores it correctly (finding 2).
- Consider adding autopilot toggle to the dropdown menu to cover more of the issue requirements (finding 4).
- Add `window.confirm` for "Reset to inbox" since it's destructive (finding 5).

### Verdict

The core implementation is solid -- the handler patterns follow existing conventions, the frontend components are clean, and the workflow state management logic is correct. The two important items (redundant write and incorrect rollback status) are straightforward fixes. The missing autopilot menu entry is a gap vs the issue spec but could be a follow-up. Requesting changes for the rollback bug (finding 2) as it could leave items in wrong states.

VERDICT: changes_requested

jasonmadigan (commented):
Confirmed: the early-return paths at lines 908-910 and 914-917 leave the workflow run in `running` status (set at line 899-901 and persisted at line 902) without rolling it back to `errored`. Only the dispatch failure path (lines 935-941) does the rollback. This is a bug.

Here's the full review:

---

## Review: PR #142 -- UI controls for work item state management

### CI Status

**Pending** -- `build-and-test` was still running. The tmux test failure that shows locally is pre-existing (commit `a20f464`, untouched by this PR).

### Summary

Adds four new backend endpoints (`reset`, `mark-done`, `cancel-workflow`, `retry-step`) and two frontend components: a kebab menu on the panel header for item-level actions, and inline workflow controls (cancel, retry, dismiss) on the workflow progress bar. The second commit addresses earlier review feedback (redundant write, rollback bug, missing autopilot cancel, confirm dialog, escape handler).

### Findings

**Important:**

1. **`handleRetryStep` -- orphaned running state on early return** (`work_item_handlers.go:897-917`). The workflow run is reset to `running` and persisted at line 902, but the subsequent checks (step out of range at 908, skill not found at 914) return errors without rolling the run back to `errored`. This leaves the run stuck in `running` with no session driving it. The dispatch failure at line 935 correctly rolls back, but these two earlier exits don't. Fix: move the validation checks (step range, skill lookup) *before* the `UpdateWorkflowRun` call, or add rollback to those error paths.

**Minor:**

2. **"Move to ready" uses the generic PATCH endpoint** (`Panel.tsx:598-599`). Unlike reset and mark-done, this doesn't cancel active sessions or workflows. The UI hides it when `status === 'in_progress'`, which helps, but items in `ready` could still have leftover sessions from previous dispatches. Low risk in practice since sessions are tied to workflow state, but it's inconsistent with the other actions that do cleanup.

3. **No test coverage for the new endpoints.** The four new handlers have no corresponding tests. Given the state machine complexity (especially retry with its rollback logic), even basic handler tests would catch issues like finding 1.

### Suggestions

- Move step/skill validation before the `UpdateWorkflowRun` call in `handleRetryStep`, or add rollback on those error paths. This is the only blocking issue.
- Consider adding handler tests for at least `handleRetryStep` and `handleResetItem` -- they have the most complex state transitions.

### Verdict

The second commit addressed the previous round of feedback well. The implementation is clean and follows existing patterns. One remaining bug: the retry handler can leave a workflow run orphaned in `running` status on validation failures. Requesting changes for that.

VERDICT: changes_requested

PR #143, opened 2026-04-14
Panel doesn't follow new sessions when workflow advances to next step

### LINKED ISSUE #137: Panel doesn't follow new sessions when workflow advances to next step
## Bug

When a workflow advances from one skill to the next (e.g. review-pr -> address-feedback), a new session is created for the same work item. The panel's live output stays on the old session and stops updating.

## Root cause

The useEffect that tracks activeSessionId and displaySessionId only re-runs when selectedId or items changes. A new session for the same work item triggers neither.

## Fix

Detect new session IDs from existing WebSocket agent_state events. When a new session ID appears for the selected work item, update the display refs. No new API calls needed.

jasonmadigan:
## Bug

When a workflow advances from one skill to the next (e.g. review-pr -> address-feedback), a new session is created for the same work item. The panel's live output stays on the old session and stops updating.

## Root cause

The useEffect that tracks activeSessionId and displaySessionId only re-runs when selectedId or items changes. A new session for the same work item triggers neither.

## Fix

Detect new session IDs from existing WebSocket agent_state events. When a new session ID appears for the selected work item, update the display refs. No new API calls needed.

Closes jasonmadigan/clawdio#137


jasonmadigan:
[clawdio review]

---

## CI Status

Passing (build-and-test: 33s).

## Summary

Fixes the panel not following new sessions when a workflow advances to the next skill step. Extracts session-ref logic into a shared `updateSessionRefs` callback and uses it both on initial load and when the WebSocket delivers an `agent_state` event for an unknown session ID. Single file change, +35/-19 in `ui/src/App.tsx`.

## Findings

### Important

1. **Stale closure in `updateSessionRefs`** (`App.tsx:85-104`). The `useCallback` depends on `[liveOutputs]`, which means it recreates on every output update. More importantly, the WebSocket handler at line 150 captures a potentially stale version of `updateSessionRefs` since `useWebSocket` doesn't list it as a dependency. This means the `!liveOutputs[target.id]` check inside the callback may read an old snapshot of `liveOutputs`, leading to unnecessary refetches of persisted output. Not a correctness bug — the `setLiveOutputs` guard (line 98: `if (prev[target.id] && prev[target.id].length > 0) return prev`) prevents data loss — but it's wasted work. Consider using a ref for `liveOutputs` in the check, or moving the `liveOutputs` guard entirely into the setter (which it already is), removing the outer check and the dependency.

### Minor

2. **Redundant `|| []` at line 117**. `updateSessionRefs` already handles the `sessions || []` normalisation internally (line 86). The caller passes `sessions || []` which double-guards. Harmless but noisy.

3. **Potential chatty API calls** (`App.tsx:146-153`). Every `agent_state` event for an unknown session triggers a `listSessions` call. If multiple events arrive in quick succession for the same new session (e.g. rapid state transitions on startup), this fires multiple fetches. A debounce or a check that the session was already added to `knownSessionIds` before the promise resolves would prevent redundant calls. Low severity given that `updateSessionRefs` overwrites `knownSessionIds` atomically, so the window is small.

## Suggestions

- For finding #1: simplest fix is to drop `liveOutputs` from the dependency array and rely solely on the `setLiveOutputs` guard, which already prevents overwriting existing output. The outer `!liveOutputs[target.id]` check is an optimisation to skip the fetch, but with a stale closure it doesn't reliably do that anyway.
- For finding #3: adding the new `sid` to `knownSessionIds.current` immediately (before the async call) would deduplicate concurrent events for the same session.

## Verdict

The approach is sound — detecting new sessions via existing WebSocket events rather than adding new polling or API calls is the right call. The core fix works: when a workflow step creates a new session, the panel picks it up and switches. The stale closure issue (#1) is worth addressing but isn't a blocking bug since the setter guard prevents data corruption. Good to merge as-is or with the minor tightening suggested above.

VERDICT: approve

PR #145, opened 2026-04-14
Timeline: fix event ordering and duplicate PR created events

### LINKED ISSUE #135: Timeline: fix event ordering and duplicate PR created events
## Bugs

**1. push_and_pr records "PR created" even when the PR already exists.**
When address-feedback pushes changes, the workflow runs push_and_pr which finds the existing PR and records a "PR created" event with the current timestamp. This makes it appear at the end of the timeline, after review and feedback cycles, which makes no sense.

Fix: only record "PR created" when a PR is actually created. When the PR already exists, record "pushed" instead (or nothing).

**2. Missing events for actions that ran before #123 landed.**
Items that went through implement -> push -> PR -> review before the workflow_events table existed have no post-action events in their timeline. The timeline shows sessions but no push/PR/merge events for historical runs.

Fix: for items with no workflow events but completed sessions, derive events from the work item state (pr_url set = PR was created at some point). This is the "falls back to derived-from-state entries" mentioned in #123's implementation, but it may not be working correctly.

## Relates to

- #123 (timeline events) -- just landed, these are bugs in the implementation

jasonmadigan:
## Bugs

**1. push_and_pr records "PR created" even when the PR already exists.**
When address-feedback pushes changes, the workflow runs push_and_pr which finds the existing PR and records a "PR created" event with the current timestamp. This makes it appear at the end of the timeline, after review and feedback cycles, which makes no sense.

Fix: only record "PR created" when a PR is actually created. When the PR already exists, record "pushed" instead (or nothing).

**2. Missing events for actions that ran before #123 landed.**
Items that went through implement -> push -> PR -> review before the workflow_events table existed have no post-action events in their timeline. The timeline shows sessions but no push/PR/merge events for historical runs.

Fix: for items with no workflow events but completed sessions, derive events from the work item state (pr_url set = PR was created at some point). This is the "falls back to derived-from-state entries" mentioned in #123's implementation, but it may not be working correctly.

## Relates to

- #123 (timeline events) -- just landed, these are bugs in the implementation

Closes jasonmadigan/clawdio#135


jasonmadigan:
[clawdio review]

Only one call site. Good.

Here's my review:

---

## CI Status

Passing (build-and-test: 39s).

## Summary

Fixes two timeline bugs: (1) `push_and_pr` handler was unconditionally recording a "PR created" event even when the PR already existed, causing it to appear out of order at the end of the timeline; (2) the frontend fallback logic for pre-workflow-events items was checking individual event types rather than gating on whether any real events exist at all, which could produce confusing hybrid timelines.

## Findings

**Minor: dead variable `eventTypes`**

`ui/src/components/Timeline.tsx:72` -- `const eventTypes = new Set(events.map(e => e.event_type))` is now unused. The old code used it in the fallback conditions, but the new `events.length === 0` guard replaces both uses. This will likely trigger a lint/TS warning for unused variables. Should be removed.

## Suggestions

Remove the dead `eventTypes` declaration on line 72. One-liner.

## Verdict

The Go-side fix is clean and correct -- gating `pr_created` on `!existed` is exactly right. The frontend fallback rewrite is also a better design: the `events.length === 0` guard means the fallback only kicks in for genuinely legacy items with no workflow events, avoiding the old issue where partial real events got mixed with synthetic ones. The only nit is the unused variable.

Approve with a note to clean up `eventTypes`.

VERDICT: approve

Issue #121, opened 2026-04-14
Convert hardcoded workflow actions to action contracts in skills/


jasonmadigan:
## Problem

Workflow actions (push, merge, create_pr, wait_for_ci, etc.) are hardcoded in a Go switch statement in `workflow/engine.go`. They can't be composed in the workflow editor, configured per-workflow, or extended without changing Go code.

Skills declared `type: action` in #87's design but were never implemented. The workflow engine still has:

```go
switch action {
case "push_and_pr":
    return executePushAndPR(...)
case "push":
    return executePush(...)
case "merge":
    return executeMerge(...)
}
```

## What should change

Each action becomes a contract file in `skills/`:

```yaml
# skills/actions/merge.yaml
name: merge
type: action
description: mark PR ready, wait for CI, squash merge
inputs:
  pr_url: { from: work_item.pr_url, required: true }
preconditions:
  - ci_passing
  - pr_ready
outputs:
  result:
    type: enum
    values: [merged, failed, conflict, ci_failed]
transitions:
  merged: done
  failed: pause
  conflict: pause
  ci_failed: pause
```

```yaml
# skills/actions/push.yaml
name: push
type: action
description: push committed changes to remote
inputs:
  worktree_path: { from: session.worktree_path, required: true }
preconditions:
  - has_commits
outputs:
  result:
    type: enum
    values: [pushed, nothing_to_push, failed]
transitions:
  pushed: next
  nothing_to_push: next
  failed: pause
```

```yaml
# skills/actions/create_pr.yaml
name: create_pr
type: action
description: create a pull request for the current branch
inputs:
  worktree_path: { from: session.worktree_path, required: true }
outputs:
  result:
    type: enum
    values: [created, exists, failed]
transitions:
  created: next
  exists: next
  failed: pause
```

## Full list of actions to convert

| Action | Current location | Inputs | Key outputs |
|-|-|-|-|
| push | engine.go:executePush | worktree_path | pushed, nothing_to_push |
| push_and_pr | engine.go:executePushAndPR | worktree_path | pr_created, pushed |
| create_pr | (part of push_and_pr) | worktree_path, title, body | created, exists |
| merge | engine.go:executeMerge | pr_url | merged, ci_failed, conflict |
| mark_ready | (part of merge) | pr_url | ready, already_ready |
| wait_for_ci | (part of merge) | pr_url | passing, failing, timeout |
| approve_pr | actions.go | pr_url | approved, cannot_approve |
[...]

PR #146, opened 2026-04-14
Action contracts should support retry for transient failures

### LINKED ISSUE #131: Action contracts should support retry for transient failures

jasonmadigan:
Workflow post-actions (push, create_pr, merge) fail permanently on transient errors like GitHub API timeouts or DNS blips. The workflow errors and the item gets stuck.

Observed: push succeeded but create_pr failed with "error connecting to api.github.com". The workflow errored with no retry. The implementation was complete but never got a PR.

Action contracts (#121) should declare retry behaviour:

\`\`\`yaml
name: create_pr
type: action
retry:
  max_attempts: 3
  backoff: 10s
  retryable_errors: [connection, timeout, 5xx]
\`\`\`

The workflow engine retries transient failures automatically, only erroring on persistent failures or after max attempts.

Relates to #121 (action contracts).

Closes jasonmadigan/clawdio#131


jasonmadigan:
[clawdio review]

Good. The refactor of `ExecutePostAction` is clean -- previously it had two paths that both called `executeActionByName` (contract path and fallback path). Now it unifies through `executeWithRetry`, which does the right thing when `retry` is nil (single attempt, immediate error propagation). Behaviour is preserved for actions without retry configs.

Here's my review:

---

## CI Status

Build-and-test: **pending** (still running at time of review).
Local test run: `workflow` and `skill` packages both pass.

## Summary

Adds retry support to action contracts. Actions like `push`, `create_pr`, and `merge` can now declare `retry:` in their YAML frontmatter to automatically retry transient failures (connection errors, timeouts, 5xx). The workflow engine retries with backoff, respects context cancellation, and records retry events for observability.

## Findings

### Important

1. **`parseIntOr` doesn't handle overflow or negative signs** (`frontmatter.go:19-31`). The hand-rolled int parser will silently overflow on very large values and won't reject `-3` (the `-` hits the `return fallback` path, which is fine, but a leading `+` would also hit fallback). This is low-risk since the inputs are controlled YAML values, but `strconv.Atoi` with a fallback would be simpler and more correct. Minor point -- not blocking.

2. **`parseDurationSeconds` only handles `s` suffix** (`frontmatter.go:35-39`). If someone writes `backoff: 10m` or `backoff: 1m30s`, they silently get the fallback value (0 with current defaults, then 10 via `effectiveRetry`). Worth a comment in the YAML or the code to make `s`-only clear, or consider supporting `m`. Not blocking since the action YAMLs are all `10s`.

### Minor

3. **Empty `retryable_errors` means "retry everything"** (`engine.go:268-270`). The logic is: if patterns list is non-empty and the error doesn't match, bail. If patterns list is empty, retry all errors. This is documented in the comment and consistent, but it's a footgun -- omitting the field entirely means every failure gets retried, including auth errors or "repository not found". Consider whether the default for empty patterns should be "retry nothing" instead. The current action YAMLs all declare explicit patterns so this doesn't bite today.

4. **`retryTestTarget` struct is unused** (`retry_test.go:83-89`). It's defined but never referenced -- looks like leftover scaffolding from an earlier test approach.

5. **`callCount` in `TestExecuteWithRetryNoRetryConfig` is unused** (`retry_test.go:126`). Declared and never incremented, along with `origFn`. Dead code in test.

## Suggestions

- Swap `parseIntOr` for `strconv.Atoi` with fallback. Less code, handles edge cases.
- Add a brief comment to the YAML files or docs noting that `backoff` only supports seconds (e.g. `# seconds only, e.g. 10s or 10`).
- Clean up unused test scaffolding (`retryTestTarget`, `callCount`, `origFn`).

## Verdict

Clean, well-structured addition. The retry logic is correct, context cancellation is handled properly, and test coverage is solid. The issues above are all minor cleanup -- nothing that affects correctness. Approve.

VERDICT: approve

Issue #113, opened 2026-04-14
Support multiple PRs per issue: track PR history, not just current PR


jasonmadigan:
## Problem

\`pr_url\` on a work item is a single string. But issues can have multiple PRs over their lifecycle:

- First attempt merged, issue reopened, second PR needed
- Work split across multiple PRs (backend + frontend)
- Failed PR closed, new one created
- Address-feedback creates commits on an existing PR, but a full redo needs a new one

The current single \`pr_url\` field causes confusion when the state changes -- skills, post-actions, and the workflow engine all assume one PR per issue.

## What should change

**Option A: pr_url is the active PR, history via child items**

Keep \`pr_url\` as a single field meaning "the current active PR". When a PR is merged/closed and the issue continues, clear \`pr_url\`. The historical PRs are already tracked as child work items (\`[PR] ...\` items with \`parent_id\`).

- Simplest change, works with existing child item model
- Skills always see the current PR via \`pr_url\`
- History visible through linked child items on the board

**Option B: pr_urls as a JSON array**

Replace \`pr_url TEXT\` with \`pr_urls TEXT\` (JSON array). The workflow engine works with the last entry. Skills can see all PRs.

- More complex queries
- Doesn't add much over option A since child items already exist

**Recommendation: Option A** with these rules:
- \`pr_url\` = the currently active, open PR (or empty)
- When a PR is merged and the issue stays open, clear \`pr_url\`
- When a PR is closed without merge, clear \`pr_url\`
- The workflow engine checks \`pr_url\` before creating a new PR
- Child PR work items (\`[PR] ...\`) provide the full history

## Skills impact

- review-pr: already gets \`pr_url\` from inputs, no change needed
- address-feedback: reads comments from the PR in \`pr_url\`, no change
- implement-issue: doesn't care about PRs, no change
- push_and_pr action: checks \`pr_url\` state before creating -- if empty or merged, creates new PR

## Relates to

- #112 (push_and_pr should ignore merged PRs)
- #109 (reopen -> inbox should clear pr_url)
- #100 (Ship It resume uses pr_url to determine state)
- #110 (DB as source of truth -- pr_state tracking)

Issue #52, opened 2026-04-12
Grouped board cards: nest linked PRs under their parent issue


jasonmadigan:
Currently every work item is a peer on the board. PRs linked to issues appear as separate cards, and the parent/child relationship is only visible in the detail panel. This makes it hard to see the full picture of a piece of work.

## Desired behaviour

When a parent issue has linked children (PRs, sub-tasks), auto-expand them as nested items inside the parent's card. No manual toggle -- if children exist, they show.

- Parent issue card renders its children inline, nested underneath
- Each child shows its own status badge (so you can see "issue is in progress, PR needs review" at a glance)
- The group lives in the **parent's column** -- children don't pull the group to a different column
- Linked PRs no longer appear as top-level cards (avoids duplicates)
- Orphan PRs (no parent) still appear as standalone cards

## Why parent drives column position

Letting children drive position (e.g. PR in NEEDS REVIEW pulls the whole group there) feels surprising. Keeping the group in the parent's column is predictable. The child status badges give you the information without the column jumping.

## Implementation notes

- `App.tsx` already computes `childrenMap` and `parentMap` from `parent_id` relationships
- Board rendering needs to filter out items that have a `parent_id` (they render inside their parent's card instead)
- `Card.tsx` needs a nested children section that renders when `childrenMap[item.id]` is non-empty
- Each nested child needs a compact representation: status badge, type, ref, title

PR #147, opened 2026-04-14
Issue-centric UX: PR actions and status on the parent issue panel, not separate work items

### LINKED ISSUE #114: Issue-centric UX: PR actions and status on the parent issue panel, not separate work items

jasonmadigan:
## Problem

When working on an issue, the workflow creates a PR as a separate work item. To action the PR (review, merge, see CI status), you need to click away from the issue to the PR work item. This is a jarring context switch -- the issue is what you care about, the PR is just an artefact of working on it.

## Desired behaviour

The issue panel is the home for the full lifecycle:

- Issue description and timeline
- Linked PR section showing: PR title, status (draft/open/merged), CI status, review verdict
- PR action buttons directly on the issue panel: push, create PR, review, approve, merge
- Agent output from all sessions (implement, review, address-feedback) in one timeline
- No need to switch to the PR work item to do anything

## When PR work items still appear standalone

PR work items should only show as standalone board cards when they're orphans:
- An external PR assigned to you for review (no parent issue in your board)
- A PR from another repo with no linked issue

When a PR has a parent issue, it renders as a nested element under the parent (already done via #52 grouped cards), and all actions live on the parent's panel.

## What changes

**Panel.tsx:**
- When the selected item has a linked PR (via childrenMap or pr_url), show PR status and actions inline
- PR actions (push, review, merge) work on the pr_url, not the current item's source_ref
- Timeline shows sessions from both the issue and its linked PR work items

**Board.tsx:**
- PR work items with a parent_id are already hidden from the top-level board (#52)
- No change needed here

**Workflow engine:**
- Already works on the issue work item, not the PR work item
- pr_url on the issue drives the review/merge steps

## Relates to

- #52 (grouped board cards, merged) -- visual nesting is done, but actions are still on the PR panel
- #113 (multiple PRs per issue, closed) -- the issue panel shows the active PR
- #110 (DB source of truth) -- PR state tracked on the issue work item

Closes jasonmadigan/clawdio#114


jasonmadigan:
[clawdio review]

Good, CI is passing. Now I have everything needed.

---

## CI Status

Passing (build-and-test: 39s).

## Summary

Adds inline PR status display and PR action buttons (approve, merge) to the issue panel, and merges child PR sessions into the issue's timeline. 3 files changed, +150/-4 lines. Pure frontend change.

## Findings

### Important

1. **`IssuePRActions` sends approve/merge to the issue work item ID, but `reviewContext` on the server requires `pr_url` on that item** (`Panel.tsx:938-949`). When the PR URL only lives on the *child* work item (not the issue itself), the server's `reviewContext` will fail with `"no PR number found"` because `item.PRURL` on the issue is empty and `item.SourceRef` is an issue reference (e.g. `#114`), not a PR. The frontend resolves `prDisplayURL` from the child for display, but `handleApprovePR`/`handleMergePR` call `api.approvePR(item.id)` with the parent issue's ID, not the child PR's ID.

   This is the core use case described in the issue ("PR actions on the parent issue panel"), and it will error whenever the PR URL hasn't been propagated to the issue's own `pr_url` field. If the workflow engine already sets `pr_url` on the parent issue, this works -- but the fallback path through `activePRChild` exists specifically because that isn't always the case. Either the API calls should pass the child PR's item ID, or the server's `reviewContext` should look up children when the item has no PR URL.

2. **`useEffect` dependency lint suppression** (`Panel.tsx:422`). The `eslint-disable-line react-hooks/exhaustive-deps` hides that `linkedChildren` (the actual array) isn't in the dependency list -- only `linkedChildren.length` is. If the children change identity but keep the same count (e.g. one PR is replaced by another), the effect won't re-run and `childSessions` will be stale. Using `linkedChildren.length` as a proxy is fragile. Consider deriving a stable key like `linkedChildren.map(c => c.id).join(',')` instead.

### Minor

3. **No deduplication of merged sessions** (`Timeline.tsx:166-169`). If a session belongs to both `sessions` and `childSessions` (e.g. because the session is listed under both parent and child), `buildEntries` will produce duplicate timeline entries. Unlikely given the current data model, but worth a defensive filter by session ID.

4. **`prState` defaults to empty string, then `|| 'open'` in the badge** (`Panel.tsx:772`). When `prState` is empty (no PR state yet), the badge renders "open" -- which may be misleading if the PR hasn't actually been created yet. The guard `hasPR` should prevent this, but it's worth noting the assumption: `hasPR` being true always means `prState` is meaningful.

## Suggestions

- For finding #1: the simplest fix is to pass the child PR's item ID to the approve/merge API calls when the PR URL comes from a child. Something like:
  ```ts
  const prActionItemId = item.pr_url ? item.id : activePRChild?.id
  ```
  Then use `prActionItemId` in `handleApprovePR` and `handleMergePR`. Alternatively, teach the server to resolve the PR from child items.

- For finding #2: replace `linkedChildren.length` with a derived string key:
  ```ts
  const childKey = linkedChildren.map(c => c.id).join(',')
  ```
  Use `childKey` in the dependency array instead.

## Verdict

Finding #1 is a functional bug for the primary use case. When the PR URL lives only on the child work item, the approve and merge buttons will fail server-side. This needs to be addressed before merging.

VERDICT: changes_requested

Issue #64, opened 2026-04-13
Dockerfile and container image for easy onboarding


jasonmadigan:
Create a Docker image with everything needed to run clawdio. Goal: try it in 60 seconds.

## Usage

\`\`\`bash
docker run -it \\
  -e ANTHROPIC_API_KEY=sk-... \\
  -v ~/.config/gh:/home/clawdio/.config/gh:ro \\
  -v clawdio-data:/data \\
  -p 8077:8077 \\
  ghcr.io/jasonmadigan/clawdio:latest
\`\`\`

Open http://localhost:8077, configure repos, start dispatching.

## What's in the image

- clawdio binary (with embedded frontend, depends on #63)
- claude CLI
- gh CLI (authenticated via mounted config)
- tmux (for session management, relates to #58)
- git
- sqlite

## Credentials

- **Anthropic API key:** env var \`ANTHROPIC_API_KEY\`
- **GitHub auth:** mount \`~/.config/gh\` read-only. The gh CLI picks it up automatically
- **Claude Code config:** optionally mount \`~/.claude\` for skills, CLAUDE.md, etc.

## Persistence

- \`/data\` volume for SQLite DB, worktrees, and config
- Named volume survives container restarts
- Config file at \`/data/config.json\` (or generated from env vars on first run)

## Image build

Multi-stage Dockerfile:
1. **ui-build stage:** Node, \`npm ci && npm run build\`
2. **go-build stage:** Go, copy ui/dist, \`go build\`
3. **runtime stage:** minimal base (debian-slim or alpine), install claude CLI, gh CLI, tmux, git, copy binary

## Future

- CI to build and push to ghcr.io on tag/release
- docker-compose.yml for local dev with volume mounts
- ARM64 support for running on cloud instances

## Depends on

- #63 (embedded frontend) -- single binary, no Node at runtime
- #58 (tmux wrapping) -- nice to have for peek/persistence inside the container

PR #148, opened 2026-04-14
gh CLI authentication in Docker containers

### LINKED ISSUE #116: gh CLI authentication in Docker containers

jasonmadigan:
## Problem

The Docker container needs gh CLI authenticated to poll GitHub and create PRs. Current approach: mount `~/.config/gh` from the host. This works if the user is already logged in on their machine, but fails if they're not.

### 1. Mount host credentials (current)

```bash
docker run -v ~/.config/gh:/home/clawdio/.config/gh:ro ...
```

Pros: zero setup if already logged in. No secrets in env vars.
Cons: fails for users who aren't logged in. Read-only mount means the container can't refresh tokens.

### 2. GitHub token via env var

```bash
docker run -e GH_TOKEN=ghp_... ...
```

Pros: simple, works everywhere, no mount needed.
Cons: PAT tokens have broad scope. User has to create and manage the token.

### 3. Auth flow via the UI

When gh isn't authenticated, show a setup step in the UI:
- "GitHub not connected" banner
- Click to start `gh auth login --web` flow
- UI shows the device code and link
- User authenticates in their browser
- Credentials saved inside the container volume

Pros: best UX, no pre-setup needed.
Cons: needs the UI to interact with the gh auth flow. Device code flow needs terminal-like interaction.

### 4. Read-write mount with fallback

Mount `~/.config/gh` read-write. If not authenticated, the container runs `gh auth login` on first start and the creds persist on the host.

```bash
docker run -v ~/.config/gh:/home/clawdio/.config/gh ...
```

Pros: works for both logged-in and not-logged-in users. Creds persist.
Cons: modifying host filesystem from a container feels wrong.

## Recommendation

Support all four, in priority order:
1. `GH_TOKEN` env var (simplest, works in CI/headless)
2. Mount `~/.config/gh:ro` (zero setup for logged-in users)
3. Auth flow in UI (best experience for new users)
4. `docker exec -it <container> gh auth login` as escape hatch

The docker-entrypoint.sh should check for auth on startup and print clear instructions if not authenticated.

## Relates to

- #64 (Dockerfile, merged) -- Docker quickstart guide

Closes jasonmadigan/clawdio#116


jasonmadigan:
[clawdio review]

Good. Now I have enough context for the review.

---

## CI Status

**Pending** -- `build-and-push` and `build-and-test` are still running. No failures yet but no pass confirmation either.

## Summary

Adds multi-method gh CLI authentication support for Docker containers. Three auth methods: `GH_TOKEN`/`GITHUB_TOKEN` env var, host credential mount, and manual `docker exec gh auth login`. Backend exposes a new `GET /api/github/auth-status` endpoint. Frontend shows a dismissible warning banner when unauthenticated. Entrypoint prints clear instructions on startup if auth is missing.

## Findings

### Important

1. **Test env var leakage between tests** (`internal/github/auth_test.go:10-33`). `TestCheckAuth_GHTokenEnv` and `TestCheckAuth_GitHubTokenEnv` set env vars with `os.Setenv` and defer `os.Unsetenv`, but these tests share a process. If `TestCheckAuth_NoAuth` runs while `GH_TOKEN` is still set (e.g. if Go reorders subtests), it gets a false positive. Use `t.Setenv()` instead -- it handles cleanup properly and marks the test as incompatible with `t.Parallel()`, which is the right thing here since they're mutating global state.

2. **`CheckAuth` parsing is fragile** (`internal/github/github.go:382`). `strings.Contains(outStr, "Logged in")` depends on gh CLI's human-readable output format. If gh changes the wording (e.g. localisation, version bump), this breaks silently. A more robust approach: just check the exit code of `gh auth status`. Exit 0 means authenticated, non-zero means not. The current code already checks `err != nil` on line 378 but then does a redundant string check on 382. If `gh auth status` exits 0, the user is authenticated -- the string check is unnecessary.

3. **`CheckAuth` calls `authUser` on every invocation** (`internal/github/github.go:372-373, 383-384`). This makes a live `gh api user` call each time the endpoint is hit. The frontend calls this on mount, and any page refresh triggers it again. For the token path, `gh api user` validates the token works, which is reasonable. But for host_mount, it's an extra API call that could be slow or rate-limited. Consider caching the result with a short TTL, or only fetching the user when explicitly needed.

### Minor

4. **Banner only checks auth once** (`ui/src/components/GitHubAuthBanner.tsx:9-13`). The `useEffect` has `[]` deps, so it fires on mount only. If a user authenticates via `docker exec` while the UI is open, the banner stays until page refresh. Not a blocker for this PR -- polling or WebSocket-based refresh is a follow-up concern.

5. **Silent error swallowing** (`ui/src/components/GitHubAuthBanner.tsx:12`). `.catch(() => {})` hides fetch failures completely. If the endpoint is broken, the user sees no banner and no error. This is defensible (no auth banner is better than a broken UI), but a `console.warn` would help debugging.

6. **`handleGitHubAuthStatus` placement** (`internal/server/work_item_handlers.go:1034-1037`). This handler is in `work_item_handlers.go` but it's not a work item handler. It'd fit better in its own file alongside the route group (`auth_handlers.go` or similar), matching the test file name that already exists (`auth_handlers_test.go`).

## Suggestions

- Replace `os.Setenv`/`os.Unsetenv` with `t.Setenv()` in `auth_test.go`.
- Drop the `"Logged in"` string match; rely on the exit code from `gh auth status`.
- Move `handleGitHubAuthStatus` to a file that matches `auth_handlers_test.go`.

## Verdict

Clean, well-scoped PR. The approach of checking env vars first, then falling back to `gh auth status`, is sound. Tests cover the key paths. Frontend banner is minimal and appropriate. The string-matching fragility (finding #2) is the main thing worth addressing before merge.

VERDICT: approve

Issue #93, opened 2026-04-13
Workflow builder UI: visual state machine editor and live execution view


jasonmadigan:
### 1. Live execution view (read-only)

When a workflow is running, show the state machine as a visual graph with live status:

- Each state is a node: skill steps, action steps, checks
- Edges show transitions with their conditions (on_approve, on_changes_requested)
- The current state is highlighted (pulsing/glowing)
- Completed states show their outcome (approve, complete, pushed)
- Errored states show the error
- Duration between states shown on edges
- The whole graph tells the story of the workflow at a glance

This replaces the current flat workflow progress bar (the numbered step badges) with something that actually communicates what's happening.

### 2. Workflow editor (read-write)

Build and configure workflows visually:

- Drag skills from a palette onto a canvas
- Connect them with transition edges
- Configure each step: skill, parallel count, convergence policy, timeout
- Configure each edge: condition (output value), action (target.publish, target.land)
- Add preconditions to steps (checks that must pass)
- Add actions between steps (push, create_review, wait_for_ci)
- Save as YAML (the workflow file format from #85)
- Load existing YAML workflows for editing

### What the graph looks like

```
[implement-issue] ──complete──→ [target.publish] ──success──→ [target.create_review]
                                                                      │
                                                               success│
                                                                      ▼
                  ┌─changes_requested── [review-pr] ──approve──→ [target.check_ci]
                  │                        ▲                           │
                  ▼                        │                      pass │
           [target.post_feedback]          │                           ▼
                  │                        │                    [target.land]
                  ▼                        │                           │
           [address-feedback]──complete──→ [target.publish] ───────────┘  ──→ [done]
```

With parallel steps, the graph fans out and converges:

```
                        ┌── [review-correctness] ──┐
[target.create_review] ─┼── [review-security]     ──┼── convergence(all_approve) ──→ ...
                        └── [review-style]        ──┘
```

## UI components

**Graph renderer:**
- Nodes positioned automatically (dagre/elkjs for layout)
- Or manual positioning saved in the workflow file
- Zoom and pan for large workflows
- Minimap for navigation

**Step inspector (sidebar):**
- Click a node to see its configuration
- Skill selection dropdown
- Input/output declarations (from skill contracts, #87)
- Preconditions list
- Parallel settings (count, convergence policy)
- Timeout settings

**Edge inspector:**
- Click an edge to configure the transition
- Condition (output value that triggers this edge)
- Actions to run during transition
- Labels shown on the edge

[...]

PR #149, opened 2026-04-14
Workflow editor: link to skill details and inline skill editing

### LINKED ISSUE #108: Workflow editor: link to skill details and inline skill editing
The workflow editor (#93, merged) should integrate with the skill system:

- Each skill node in the graph should link to the skill's detail view (inputs, outputs, prompt)
- Clicking a skill node opens an inspector showing the skill contract (declared inputs, outputs, transitions)
- Option to edit the skill prompt inline from the workflow editor
- The skill palette should show contract info (what outputs drive transitions)
- Validation should warn if a skill's declared outputs don't match the workflow's expected transitions

This builds on the existing workflow editor to make it a single place to understand and modify the full pipeline, not just the wiring between skills.

jasonmadigan:
The workflow editor (#93, merged) should integrate with the skill system:

- Each skill node in the graph should link to the skill's detail view (inputs, outputs, prompt)
- Clicking a skill node opens an inspector showing the skill contract (declared inputs, outputs, transitions)
- Option to edit the skill prompt inline from the workflow editor
- The skill palette should show contract info (what outputs drive transitions)
- Validation should warn if a skill's declared outputs don't match the workflow's expected transitions

This builds on the existing workflow editor to make it a single place to understand and modify the full pipeline, not just the wiring between skills.

Closes jasonmadigan/clawdio#108


jasonmadigan:
[clawdio review]

TypeScript compiles clean. Here's my review:

---

## CI Status

**Pending** -- `build-and-test` is still running. Go builds and all tests pass locally (the only failure is a pre-existing flaky test in `internal/tmux` unrelated to this PR).

## Summary

Adds contract-aware validation to the workflow engine and surfaces skill contracts in the workflow editor UI. Three main pieces:

1. **Backend**: New `ValidateWithContracts()` function that extends the existing `Validate()` to check skill output values against workflow step transitions (both directions -- unhandled outputs and bogus transitions).
2. **Frontend**: Inspector panel shows skill contracts (inputs, outputs, transitions, phase), with mismatch warnings. Inline prompt viewer/editor. Palette items show contract metadata badges.
3. **Types**: Frontend TS types for `SkillContract`, `SkillInput`, `SkillOutput`.

## Findings

### Important

1. **`validate.go:167-176` -- unhandled output validation fires even for steps with no transitions at all.** If a step has `Transitions: nil` and the skill declares outputs, this block is skipped (correct). But if `Transitions` is a non-nil empty map `{}`, the inner loop will produce warnings for every output value. This is a minor edge case, but worth noting -- a step might legitimately be a terminal step with an empty transitions map.

2. **`WorkflowBuilder.tsx:357` -- silent error swallowing on prompt save.** `catch { /* ignore */ }` means a failed save gives no feedback to the user. The `setSavingPrompt(false)` and `setShowPrompt(false)` execute regardless, so the user sees the editor close as if the save succeeded. At minimum, keep the editor open on failure. I note the existing code also swallows errors in `serializeWorkflow` and `saveWorkflow` with the same pattern, so this is consistent, but extending the pattern to a write operation (prompt save) is riskier.

### Minor

3. **`WorkflowBuilder.tsx:14` -- `hasContract` utility function uses a cast `as boolean`.** The expression `(a || b || c) as boolean` is unnecessary since `||` already returns a boolean when all operands are booleans or truthy checks on `Object.keys().length`. The Go-side `HasContract()` is cleaner -- just `len(x) > 0`.

4. **`types.ts:103` -- `SkillContract` missing `retry` field.** The Go model includes `Retry *RetryConfig`. Not needed for the current UI, but if the contract view is meant to be comprehensive, it's a gap. Low priority since retry config display isn't part of this issue's scope.

5. **`validate.go:180-193` -- only `on_` prefixed keys are validated against outputs.** Non-prefixed transition keys (without `on_`) that don't match any output value are silently ignored. This seems intentional (avoiding false positives for `goto` and custom keys), but it means a typo like `approve` instead of `on_approve` won't be caught.

6. **`styles.css` -- `.palette-item` layout change.** The diff changes `.palette-item` to `flex-direction: column` which affects all palette items, including the action buttons (`push`, `push_and_pr`, `merge`). Those don't use `PaletteSkillItem` so they won't have the meta span, but the flex-direction change still applies. Verify action buttons still look right.

## Suggestions

- For finding #2: at minimum, don't close the prompt editor on error. Consider a brief inline error state or keeping the textarea visible.
- For finding #1: add a guard `if len(step.Transitions) == 0 { continue }` before the unhandled-output loop, or document that empty-map-with-no-keys is not a valid state.
- The test coverage for the new validation logic is solid -- four tests covering unhandled outputs, all-handled, bogus transitions, and no-contract.

## Verdict

This is a clean, well-scoped implementation. The validation logic is sound with good test coverage. The UI additions are sensible. The two important findings are real but not blockers -- the silent error swallowing is the most actionable. Approving.

```
VERDICT: approve
```

Issue #87, opened 2026-04-13
Skills as contracts: declared inputs, outputs, preconditions, and transitions


jasonmadigan:
## The problem

Skills are opaque prompts. The orchestrator has no idea what a skill needs to run, what it produces, or how to verify it worked. All the intelligence about skill behaviour lives in hardcoded Go:

- `parseReviewVerdict` string-matches agent output to guess the review outcome
- `ProcessPostSession` / `ProcessReviewPostSession` / `ProcessFeedbackPostSession` branch on skill names
- Post-actions are declared in YAML frontmatter but their semantics are hardcoded
- The workflow engine knows that review-pr produces a verdict and address-feedback needs review comments, but only because humans wrote the wiring

This makes skills non-composable. Adding a new skill requires changing Go code to handle its outputs and transitions.

## Skills as contracts

A skill should declare its interface the way a function declares its signature:

```yaml
---
name: review-pr
description: Review a pull request for correctness, style, and completeness

inputs:
  pr_url:
    from: work_item.pr_url
    required: true
  worktree_path:
    from: session.worktree_path
    required: true

preconditions:
  - check: pr_exists
    description: a review artifact must exist before we can review it
  - check: worktree_exists
    description: agent needs a workspace to read the code

outputs:
  verdict:
    type: enum
    values: [approve, changes_requested]
    parse: |
      look for a "## Verdict" section near the end of the output.
      "approve" if the section contains "approve" without "request changes".
      "changes_requested" if it contains "request changes" or "changes requested".
    default: changes_requested

transitions:
  approve: target.check_ci
  changes_requested: target.post_review_feedback

post_complete:
  - target.post_review_feedback   # always post the review text, regardless of verdict
---

[skill prompt content here]
```

### What each section does

**inputs** -- what data the skill needs in the dispatch prompt. The orchestrator resolves these from work item fields, session state, or target adapter queries. If a required input is missing, the skill can't be dispatched (fail early, not mid-session).

**preconditions** -- checks that must pass before dispatch. These map to `Check` primitives (see #86). If a precondition fails, the workflow engine knows to skip or pause rather than dispatch a session that will waste time and tokens.

**outputs** -- what the skill produces. The orchestrator parses the agent's output (the session summary) to extract structured values. The `parse` field describes how to extract the value -- this replaces the hardcoded `parseReviewVerdict` function with a per-skill, declarative extraction rule.

[...]

Issue #76, opened 2026-04-13
Autopilot prioritisation: triage skill to order dispatch queue


jasonmadigan:
Autopilot currently dispatches matching inbox items in arbitrary order (whatever the DB returns). It should use a configurable triage step to prioritise which items to work on first.

## How it works

When autopilot has multiple matching items and limited slots (max_concurrent), it needs to decide which items to dispatch first. A triage skill runs a quick classification pass over the pending items and returns a ranked order.

## Approach

1. **Triage skill** -- a dispatch skill (e.g. \`skills/triage.md\`) that receives a batch of items and returns them ranked by priority. Could use heuristics (labels, age, repo, type) or a quick Claude call for smarter classification.

2. **Integration with autopilot** -- before dispatching, autopilot passes the matched items through the triage skill. The skill returns an ordered list. Autopilot dispatches from the top.

3. **Configurable** -- the triage skill name is set in the autopilot config:
\`\`\`json
"autopilot": {
  "triage_skill": "triage",
  ...
}
\`\`\`
If not set, items dispatch in arrival order (current behaviour).

## Triage signals

The skill could consider:
- Labels (bug vs feature vs docs)
- Age (older items first, or newer?)
- Repo priority (some repos matter more)
- Item type (issues vs PR reviews -- reviews might be time-sensitive)
- Size hints (PR diff size, issue description length)
- Dependencies (does this unblock other items?)

## Relates to

- #24 (auto-triage: classify and pre-select skills) -- this is the prioritisation half of that issue
- #23 (autopilot) -- this enhances autopilot's dispatch ordering

jasonmadigan:
With skill contracts (#87) landed, the triage skill can declare a `priority` output as an enum. The autopilot engine reads the output and dispatches in priority order. No additional framework needed beyond what contracts already provide.

PR #150, opened 2026-04-14
Auto-triage: classify incoming items and pre-select skills

### LINKED ISSUE #24: Auto-triage: classify incoming items and pre-select skills
## Narrowed scope

The broad triage vision (classify content, pick skills) is now covered by:
- Autopilot rules (#23, merged) -- type-to-skill mapping and auto-dispatch
- Triage prioritisation (#76, merged) -- ordering the dispatch queue

The remaining unique value of this issue is the **UI layer**: showing triage results on work item cards before manual dispatch.

## What to build

A suggested skill badge on each work item card:
- Shows the recommended skill with brief rationale (e.g. "review-pr -- PR assigned for review")
- Editable before dispatch -- click to change the skill
- Populated by autopilot rules initially, upgradable to content-aware classification later

This gives users visibility into what would happen on dispatch without needing to open the dispatch modal.

## Update: skill contracts landed (#87)

With contracts, triage can be expressed as a skill with declared outputs:

```yaml
name: triage
outputs:
  priority:
    type: enum
    values: [critical, high, normal, low]
    parse: "PRIORITY:\\s*(critical|high|normal|low)"
  suggested_skill:
    type: string
    parse: "SKILL:\\s*(\\S+)"
```

The triage result feeds into autopilot dispatch ordering (#76) and the UI skill suggestion badge.

jasonmadigan:
## Narrowed scope

The broad triage vision (classify content, pick skills) is now covered by:
- Autopilot rules (#23, merged) -- type-to-skill mapping and auto-dispatch
- Triage prioritisation (#76, merged) -- ordering the dispatch queue

The remaining unique value of this issue is the **UI layer**: showing triage results on work item cards before manual dispatch.

## What to build

A suggested skill badge on each work item card:
- Shows the recommended skill with brief rationale (e.g. "review-pr -- PR assigned for review")
- Editable before dispatch -- click to change the skill
- Populated by autopilot rules initially, upgradable to content-aware classification later

This gives users visibility into what would happen on dispatch without needing to open the dispatch modal.

## Update: skill contracts landed (#87)

With contracts, triage can be expressed as a skill with declared outputs:

```yaml
name: triage
outputs:
  priority:
    type: enum
    values: [critical, high, normal, low]
    parse: "PRIORITY:\\s*(critical|high|normal|low)"
  suggested_skill:
    type: string
    parse: "SKILL:\\s*(\\S+)"
```

The triage result feeds into autopilot dispatch ordering (#76) and the UI skill suggestion badge.

Closes jasonmadigan/clawdio#24


jasonmadigan:
[clawdio review]

Here's the review:

---

## CI Status

**Passing** -- `build-and-test` green.

## Summary

Adds a suggested skill badge to work item cards. Backend computes suggestions by checking autopilot rules, then type mapping, then default skill -- frontend fetches suggestions and renders a clickable badge that pre-populates the dispatch modal. 10 files changed, 265 additions.

## Findings

### Important

1. **Duplicated autopilot matching logic** (`internal/skill/skill.go:496-530`). The new `matchAutopilotRule` and `matchesList` are functionally identical to `matchRule`, `matchesRepo`, and `matchesType` in `internal/autopilot/autopilot.go:228-282`. The only structural difference is that `matchesList` combines both repo and type matching into one generic function, while autopilot has separate `matchesRepo` (case-insensitive) and `matchesType` (case-sensitive) helpers. This means `matchesList` is actually case-insensitive for item types, which is a subtle behavioural divergence from autopilot -- a suggestion could match where autopilot wouldn't. Worth either sharing the matching logic or at least making case sensitivity consistent.

2. **Duplicated `AutopilotRule` type** (`internal/skill/skill.go:491-495`). The comment says "mirrors config.AutopilotRule to avoid circular imports" but it drops the `AutoPR` field. The conversion in the handler (`work_item_handlers.go:37-41`) manually copies fields. If `config.AutopilotRule` gains new fields (e.g. a `Priority` field), the skill package won't see them. Consider moving the shared matching logic to a small shared package, or having the handler pass `config.AutopilotRule` directly with the skill package importing config (which shouldn't cause a cycle since skill doesn't import autopilot).

3. **`Suggest` doesn't check autopilot rule skill exists** (`internal/skill/skill.go:470-472`). The type mapping path calls `m.Get(name)` to verify the skill exists before suggesting it. The autopilot rule path returns the suggestion without verifying. If a rule references a skill that doesn't exist, the user gets a badge they can't dispatch.

### Minor

4. **Silent error swallowing on suggestion fetch** (`ui/src/App.tsx:63`). The `.catch(() => {})` means the user gets no indication if the suggestions endpoint is broken. Fine for a non-critical feature, but worth a console.warn at minimum for debuggability.

5. **Suggestion endpoint iterates all work items** (`work_item_handlers.go:30`). `ListWorkItems("")` fetches everything, then filters to inbox/ready. Not a problem at current scale, but the query could filter by status in SQL if the list grows.

## Suggestions

- Extract the autopilot matching logic into a shared package (e.g. `internal/autopilot/match.go` with exported functions) that both `autopilot.Engine` and `skill.Manager` can use. Eliminates the duplicated type and matching functions, and prevents case-sensitivity divergence.
- Add a `m.Get(rule.Skill)` check in the autopilot rule branch of `Suggest`, matching what the type mapping branch does.

## Verdict

The feature works as described and the tests cover the key paths well. The duplication and case-sensitivity divergence are worth addressing but aren't blocking -- they're improvement candidates, not bugs in the current feature scope.

VERDICT: approve

PR #152, opened 2026-04-14
Add a refresh/re-poll button to trigger GitHub import on demand

### LINKED ISSUE #144: Add a refresh/re-poll button to trigger GitHub import on demand
New work items only appear when the poller runs (every 30s-5m depending on config) or on page refresh. There's no way to manually trigger a poll from the UI.

## Fix

A refresh button on the board (near the group-by toggle or in the header) that triggers an immediate poll cycle. Useful when you've just created an issue on GitHub and want it to appear immediately.

The board already has an onRefresh prop but it only reloads items from the DB -- it doesn't trigger the poller to fetch from GitHub.

Needs a new API endpoint: POST /api/poll (or similar) that triggers one poll cycle and returns the results.

## Relates to

- #132 (polling strategy) -- on-demand fetch is part of the revised polling approach

jasonmadigan:
New work items only appear when the poller runs (every 30s-5m depending on config) or on page refresh. There's no way to manually trigger a poll from the UI.

## Fix

A refresh button on the board (near the group-by toggle or in the header) that triggers an immediate poll cycle. Useful when you've just created an issue on GitHub and want it to appear immediately.

The board already has an onRefresh prop but it only reloads items from the DB -- it doesn't trigger the poller to fetch from GitHub.

Needs a new API endpoint: POST /api/poll (or similar) that triggers one poll cycle and returns the results.

## Relates to

- #132 (polling strategy) -- on-demand fetch is part of the revised polling approach

Closes jasonmadigan/clawdio#144


jasonmadigan:
[clawdio review]

CI passed. Here's the review:

---

## CI Status

Passing (build-and-test completed successfully).

## Summary

Adds a "Refresh" button to the `web/` (vanilla JS) UI header that triggers `POST /api/poll` to run an on-demand GitHub poll cycle. The backend endpoint and its tests already exist on main -- this PR wires up the frontend only: an HTML button, a click handler with loading state, and a CSS class for the polling state.

## Findings

**Minor**

1. **Hardcoded 1500ms delay is fragile** (`web/app.js:93`). The comment says "so the backend poll completes and broadcasts", but the poll runs asynchronously on the server (`go s.pollCoord.PollOnce(...)`) and could take longer depending on how many repos are configured. The websocket broadcast should handle item updates regardless, and the explicit `loadItems()` call after the delay is redundant with the WS-driven refresh. Consider either: (a) dropping the delay + `loadItems()` entirely and relying on the websocket broadcast, or (b) if you want the explicit reload as a fallback, making the delay configurable or at least longer. Not a blocker -- the WS fallback means items will appear regardless.

2. **No error feedback to the user** (`web/app.js:95-96`). The catch block silently swallows errors. If the server returns 503 (polling not configured), the button returns to "Refresh" with no indication anything went wrong. A brief visual hint (e.g. button text "Failed" for a second) would be better UX. Minor since the vanilla JS UI is secondary to the React UI.

3. **React UI not wired up**. The `ui/src/lib/api.ts:221` already exports a `triggerPoll` API call, but the React board doesn't have a corresponding refresh button. The issue mentions the board's `onRefresh` prop. This PR only covers the vanilla `web/` UI. Worth noting in case the React UI is the primary interface -- if so, a follow-up is needed.

## Suggestions

- Consider dropping the `setTimeout` + `loadItems()` and relying on the websocket broadcast. The poll handler fires a goroutine and returns immediately -- the WS broadcast will push item updates when they arrive, regardless of timing.

## Verdict

Clean, small change. The backend is solid (already tested), and the frontend addition is straightforward. The 1500ms delay is a minor smell but not harmful -- the WS fallback covers it. Approve.

VERDICT: approve

PR #153, opened 2026-04-14
Remove manual CI polling: let gh pr merge handle check requirements

### LINKED ISSUE #151: Remove manual CI polling: let gh pr merge handle check requirements
## Problem

The workflow engine manually polls CI status with PRChecksPass/WaitForChecks before attempting merge. This duplicates GitHub's own merge requirements logic and causes false failures when optional checks (like build-and-push) fail.

## Fix

Remove PRChecksPass and WaitForChecks entirely. Just call \`gh pr merge\`. GitHub will reject the merge if required checks haven't passed. The error message tells us exactly what's wrong.

The merge action becomes:
1. Mark PR ready (if draft)
2. Try \`gh pr merge --squash\`
3. If it fails because checks are pending, retry with backoff (#131)
4. If it fails because checks failed, error with the GitHub error message

No manual CI polling. No guessing which checks matter. GitHub is the authority.

## Also fixed

Removed \`pull_request\` trigger from docker.yml -- build-and-push only runs on main/tags, not PRs. It was failing on PRs and blocking merges as an "optional" check that we treated as required.

jasonmadigan:
## Problem

The workflow engine manually polls CI status with PRChecksPass/WaitForChecks before attempting merge. This duplicates GitHub's own merge requirements logic and causes false failures when optional checks (like build-and-push) fail.

## Fix

Remove PRChecksPass and WaitForChecks entirely. Just call \`gh pr merge\`. GitHub will reject the merge if required checks haven't passed. The error message tells us exactly what's wrong.

The merge action becomes:
1. Mark PR ready (if draft)
2. Try \`gh pr merge --squash\`
3. If it fails because checks are pending, retry with backoff (#131)
4. If it fails because checks failed, error with the GitHub error message

No manual CI polling. No guessing which checks matter. GitHub is the authority.

## Also fixed

Removed \`pull_request\` trigger from docker.yml -- build-and-push only runs on main/tags, not PRs. It was failing on PRs and blocking merges as an "optional" check that we treated as required.

Closes jasonmadigan/clawdio#151


jasonmadigan:
[clawdio review]

Only one implementation (`GitHubTarget`). No compilation risk from other implementors.

---

## CI Status

**Pending** -- `build-and-test` is still running.

## Summary

Removes manual CI polling (`PRChecksPass`, `WaitForChecks`, `CheckCI`, `WaitForCI`) and the `wait_for_ci` skill action, delegating check enforcement to GitHub via `gh pr merge`. The merge action's retry config handles transient "checks pending" failures with backoff.

## Findings

**No critical or important issues found.**

The change is clean and well-scoped:
- All references to removed code are gone (verified with grep, zero stale references).
- The `Target` interface is correctly simplified -- only one implementation (`GitHubTarget`) exists.
- The retry config in `merge.yaml` is sensible: increased from 3 to 5 attempts, backoff from 10s to 30s, and added substring patterns (`checks pending`, `required status check`) that match typical GitHub API error messages.
- The `isTransientError` function uses case-insensitive substring matching, so partial matches in `gh` CLI output will work.
- The test update is correct -- swaps `ci_passed` for `pr_ready` as the first event, which matches the new flow.
- Frontend types, labels, and styles are all updated consistently. No orphaned references.
- Net -120 lines of code. Good.

**Minor observations (not blocking):**

1. The retryable error strings `checks pending` and `required status check` are reasonable guesses at `gh pr merge` output, but the exact wording depends on the `gh` CLI version. If GitHub ever changes the phrasing, these substrings would stop matching. This is an acceptable tradeoff -- the alternative (manual polling) was worse.

2. 5 attempts x 30s backoff = ~2.5 minutes of retry ceiling. For a CI suite that takes longer, the merge would fail and the item would pause. This seems intentional per the `pause` transition on `failed`.

## Verdict

Straightforward removal of redundant logic. The approach of letting GitHub enforce its own merge requirements is correct. No bugs, no regressions, clean removal.

```
VERDICT: approve
```

PR #159, opened 2026-04-14
Trigger GitHub poll on state changes instead of relying solely on interval polling

### LINKED ISSUE #154: Trigger GitHub poll on state changes instead of relying solely on interval polling
## Problem

The poller runs on a fixed interval (30s). When a workflow completes an action (merge, push, create PR), the local state updates immediately but the GitHub state (closed issues, merged PRs, new labels) only syncs on the next poll cycle. This causes stale data for up to 30 seconds after every action.

## Fix

Trigger a poll cycle immediately after state-changing actions:
- After a PR is merged
- After a PR is created
- After a push
- After issue status sync (labels added/removed)
- After autopilot completes a workflow

The poller already has a `pollOnce()` function. Just call it after these actions instead of waiting for the next tick.

The fixed-interval poll can remain as a safety net (maybe at a slower interval like 2-5 minutes) to catch changes made outside clawdio (manual GitHub edits, external CI, other contributors).

## Relates to

- #132 (polling strategy) -- this is the event-driven complement to interval polling
- #144 (manual refresh button) -- manual trigger is the UI version of this

jasonmadigan:
## Problem

The poller runs on a fixed interval (30s). When a workflow completes an action (merge, push, create PR), the local state updates immediately but the GitHub state (closed issues, merged PRs, new labels) only syncs on the next poll cycle. This causes stale data for up to 30 seconds after every action.

## Fix

Trigger a poll cycle immediately after state-changing actions:
- After a PR is merged
- After a PR is created
- After a push
- After issue status sync (labels added/removed)
- After autopilot completes a workflow

The poller already has a `pollOnce()` function. Just call it after these actions instead of waiting for the next tick.

The fixed-interval poll can remain as a safety net (maybe at a slower interval like 2-5 minutes) to catch changes made outside clawdio (manual GitHub edits, external CI, other contributors).

## Relates to

- #132 (polling strategy) -- this is the event-driven complement to interval polling
- #144 (manual refresh button) -- manual trigger is the UI version of this

Closes jasonmadigan/clawdio#154


jasonmadigan:
[clawdio review]

Good. `TriggerPoll` mirrors the same pattern as `handlePoll` -- fire-and-forget with `context.Background()`, nil-safe check. `PollOnce` has its own mutex guard so concurrent calls just no-op.

Here's my review:

---

## CI Status

Pending -- `build-and-test` is still running. Tests pass locally.

## Summary

Adds event-driven polling by calling `TriggerPoll()` after state-changing actions (merge, push, PR creation, review, approval, request changes) so GitHub state syncs immediately rather than waiting for the next background tick. Also fixes a pre-existing compile error in `coordinator_test.go` where `BackgroundPoll` was passed as `bool` instead of `*bool`.

## Findings

No critical or important issues found. The implementation is clean and minimal.

**The approach is sound:**
- `TriggerPoll()` is nil-safe (checks `pollCoord != nil`)
- Fires asynchronously via goroutine (doesn't block HTTP handlers)
- Uses `context.Background()` correctly -- same pattern as the existing `handlePoll` handler
- `PollOnce` is already concurrency-safe with a mutex guard, so rapid successive triggers just no-op
- Rate limit checks inside `PollOnce` prevent API abuse

**Coverage of trigger points looks complete:**
- `handlePushAndPR` -- PR creation
- `handlePush` -- push
- `handlePostReview` -- review posted
- `handleRequestChanges` -- changes requested
- `handleApprovePR` -- both approve and approve-and-merge paths
- `handleMergePR` -- explicit merge
- `cmd/clawdio/main.go` -- autopilot auto-PR and post-session autonomous actions

**Test coverage:** New test for nil coordinator safety. The `coordinator_test.go` fix is a legitimate pre-existing bug (type mismatch `bool` vs `*bool`).

**Minor:** The issue mentions "after issue status sync (labels added/removed)" as a trigger point. I don't see a dedicated handler for that, but if label sync happens during the poll cycle itself, triggering another poll wouldn't help -- so this seems intentional rather than a gap.

## Suggestions

None. This is a tight, focused change that does exactly what the issue asks for.

## Verdict

Clean implementation, good test coverage, no issues found.

VERDICT: approve

PR #160, opened 2026-04-14
Bidirectional agent sessions: answer questions and approve tools

### LINKED ISSUE #9: Bidirectional agent sessions: answer questions and approve tools

jasonmadigan:
Currently sessions run with `--permission-mode bypassPermissions` (one-shot, fire and forget).

**What exists already (but isn't wired up):**
- Model types: `ControlRequest`, `AgentState` (awaiting_input, awaiting_approval)
- API endpoints: `/api/sessions/{id}/message`, `approve`, `deny`, `cancel`
- React Panel: approval box UI, input field for answering questions
- Session manager: `SendMessage()`, `ApproveToolUse()`, `DenyToolUse()` methods

**What's missing:**
- Actually using `--input-format stream-json` (caused output buffering issues in testing)
- Switching to `--permission-mode default` so the agent asks for permissions
- Testing the full bidirectional flow end-to-end
- The stdin close fix needs to be reverted when this is enabled

**Blocker from initial testing:** `--input-format stream-json` may change how claude buffers output. Needs investigation.

**When this lands:**
- Remove the non-interactive constraints from all dispatch skills (`skills/*.md`). Currently all skills say 'Do NOT wait for user input or approval' and 'Do NOT use planning skills that require approval steps'. These exist because sessions are one-shot. With bidirectional support, agents could use planning/approval workflows interactively.
- Update skill docs to note which skills benefit from interactive mode vs fire-and-forget.

## Permission mode toggle

Currently hardcoded to `--permission-mode bypassPermissions` in manager.go. Needs to be configurable at three levels:

**Global default (config.json):**
```json
"permission_mode": "default"
```
Options: `default` (agent asks for permission), `bypassPermissions` (agent runs autonomously). Default should be `default` for safety.

**UI toggle:**
Prominent toggle in the header (like the autopilot button). Visually obvious when bypass is on -- power-user mode.

**Per-dispatch override:**
Permission mode dropdown in the dispatch modal. Override the global setting for a single run.

When `default` mode is active, bidirectional comms are essential -- the agent asks permission and the UI handles approve/deny. When `bypassPermissions` is on, that UI is dormant.

## Prerequisites (now met)

- #85 (state machine) -- merged
- #86 (adapters) -- merged
- #87 (contracts) -- merged

Closes jasonmadigan/clawdio#9


jasonmadigan:
[clawdio review]

---

## CI Status

**Passing.** Build and tests succeeded.

## Summary

Adds configurable permission mode (`default` / `bypassPermissions`) at three levels: global config, UI header toggle, and per-dispatch override in both single and bulk dispatch modals. When `default` mode is active, `--input-format stream-json` is added to enable bidirectional stdin for agent questions and tool approvals. Non-interactive constraints are removed from all dispatch skills.

## Findings

### Important

1. **No validation on permission mode values.** The `PermissionMode` field is a freeform `string` everywhere -- config, API request bodies, dispatch opts, DB column. A typo like `"bypass"` or `"Default"` would be passed directly to `--permission-mode` and the `claude` CLI would likely reject it at runtime. There's no validation at any boundary: not in `config.Load()`, not in `handleDispatch`, not in `handleUpdateConfig`, not in `buildArgs`. This is the kind of thing that surfaces as a confusing runtime error. Consider at minimum a validation in `buildArgs` or `config.Load()` that rejects unknown values. Could also use a typed constant.

2. **Skills no longer declare non-interactive constraints, but interactive mode isn't conditional.** The PR removes "Do NOT wait for user input" from all skills unconditionally. But sessions dispatched with `bypassPermissions` are still non-interactive -- there's no stdin pipe. If a skill prompts the agent to use planning/approval workflows, and the session is in bypass mode, the agent could hang or fail waiting for input that will never arrive. The skill constraints should ideally be conditional on the permission mode, or the prompt construction should inject a non-interactive constraint when bypass mode is active.

### Minor

3. **`PermissionMode` stored as empty string when not overridden.** In `Dispatch()`, if no per-dispatch override is given, `permMode` stays `""` and gets stored in the session record. `buildArgs` handles the empty case with a fallback to `"default"`, but the DB row has an empty `permission_mode` which is misleading -- it doesn't reflect what actually ran. Storing the resolved value would make the session record self-documenting.

4. **Bulk dispatch duplicates the opts construction.** The pattern in `handleBulkDispatch` (lines 323-327) is copy-pasted from `handleDispatch` (lines 238-242). Minor, but a helper would keep them in sync.

5. **`getConfig` race on initial load.** `App.tsx` line 72 -- `api.getConfig().then(cfg => setPermissionMode(cfg.permission_mode || 'default'))` could silently swallow a network error. If the config fetch fails, `permissionMode` stays `'default'` regardless of what the server actually has. The `.catch(() => {})` hides this. Not a blocker, but worth noting.

## Suggestions

- Add a `ValidPermissionMode(s string) bool` helper and call it in `config.Load()` and the dispatch handlers. Two valid values, simple check.
- Re-add the non-interactive constraint to skills dynamically: in `BuildPrompt()`, when the resolved permission mode is `bypassPermissions`, append the constraint to the prompt. This keeps skills clean while protecting bypass-mode sessions from hanging.
- Store the resolved permission mode (after fallback logic) in the session record rather than the raw override value.

## Verdict

The core feature is well-structured -- config plumbing, three-level override hierarchy, DB migration, UI toggle, per-dispatch modal, bidirectional stdin gating. Tests cover the key arg-building paths. The main concern is the missing validation and the unconditional skill constraint removal, which together create a path where bypass-mode sessions could get stuck. These are worth addressing before merge.

VERDICT: changes_requested

jasonmadigan:
[clawdio review]

Good -- the static non-interactive constraints are fully removed from all skills. They're now injected dynamically by `BuildPrompt` when the resolved permission mode is `bypassPermissions`.

Here's my review:

---

## CI Status

**Pending.** Build-and-test check is still running. Tests pass locally (the only failure is a pre-existing `TestKillSessionNonExistent` in `internal/tmux` which also fails on main).

## Summary

Adds configurable permission mode at three levels: global config (`config.json`), UI header toggle, and per-dispatch override in both single and bulk dispatch modals. When `default` mode is active, `--input-format stream-json` is added to enable bidirectional stdin for agent questions and tool approvals. Non-interactive constraints are removed from skill files and instead injected dynamically into the prompt when bypass mode is active. Includes DB migration, validation, and tests.

## Previous review

The earlier review raised three points. All have been addressed in commit `1df1283`:

1. **Validation** -- `ValidPermissionMode()` now exists and is called at config load, config update, single dispatch, and bulk dispatch endpoints. Tests cover valid, invalid, and edge cases.
2. **Conditional skill constraints** -- Non-interactive constraints are removed from all skills and dynamically injected by `BuildPrompt` when the resolved permission mode is `bypassPermissions`. Tests verify both bypass and default modes.
3. **Resolved value stored** -- `Dispatch()` resolves the effective permission mode (config default + per-dispatch override + fallback) and stores the resolved value in the session record, not the raw override.

## Findings

### Minor

1. **Redundant resolution in `buildArgs`.** `buildArgs` (`manager.go:557-564`) re-resolves the permission mode from config + override, but both callers pass `sess.PermissionMode` which is already the fully resolved value from `Dispatch()`. The config fallback and empty-string fallback in `buildArgs` can never trigger. Not a bug -- just dead logic. Could simplify `buildArgs(sess model.Session)` to use `sess.PermissionMode` directly and drop the `permOverride` param.

2. **Bulk dispatch doesn't use the shared helper.** `handleDispatch` uses `dispatchOptsFromPermissionMode()` (line 256), but `handleBulkDispatch` builds opts inline (lines 341-344). The helper was introduced to de-duplicate this -- would be cleaner to use it in both places.

3. **Skill files have empty `## Constraints` headers.** After removing the non-interactive lines, `address-feedback.md`, `default.md`, `implement-issue.md`, and `review-pr.md` still have `## Constraints` headers followed by the remaining constraints, which is fine. But `resolve-conflicts.md` now has `## Constraints` with only one remaining bullet. Cosmetic only.

## Verdict

The previous review's feedback has been addressed well. The remaining points are cosmetic. The core feature -- three-level permission mode with validation, bidirectional stdin gating, and dynamic constraint injection -- is solid and well-tested.

VERDICT: approve

PR #192, opened 2026-04-15
Reduce GitHub API usage: deduplicate queries, batch status checks

### LINKED ISSUE #189: Reduce GitHub API usage: deduplicate queries, batch status checks
## Problem

Exhausting the GraphQL rate limit (5000/hour) as a single user.

## What's been fixed

- Rate limit check now covers both REST and GraphQL buckets (previously only checked REST)
- Poll interval changed from 30s to 5m
- Manual refresh button for on-demand polling

### Deduplicate polling queries
- Two configured queries hit the same 3 repos for PRs -- 6 GraphQL calls where 3 would do
- Fetch all open issues + PRs per repo once, filter client-side by assignee/author/label in Go

### Batch status checks
- `SyncStatuses` calls `gh issue view` / `gh pr view` for every tracked item individually
- Use a single `gh api graphql` query to check multiple items at once
- Or skip items that haven't changed since last poll (compare updated_at)

### Consider REST alternatives
- `gh issue list` and `gh pr list` use GraphQL internally
- `gh api /repos/{owner}/{repo}/issues` uses REST (separate 5000/hour bucket)
- Could split load across both buckets

### Long term
- When MCP architecture (#197) lands, polling could move to a GitHub MCP server
- Agent `gh` usage would also go through MCP, giving centralised rate budget visibility

jasonmadigan:
## Problem

We're exhausting the GraphQL rate limit (5000/hour) as a single user. Root causes:

1. **Duplicate queries:** Two configured queries both hit the same 3 repos for PRs -- 6 GraphQL calls where 3 would do.
2. **No client-side filtering.** Each query with different filters makes separate API calls per repo. Could fetch once and filter in Go.
3. **Status sync checks items individually.** SyncStatuses calls gh issue view for every tracked item on every poll cycle.
4. **Rate limit check only covered REST.** Fixed -- now checks both REST and GraphQL buckets.

### Deduplicate polling
- Merge queries hitting the same repos into a single fetch per repo
- Fetch all open issues + PRs once, filter client-side

### Batch status checks
- Use a single GraphQL query to check multiple items
- Or skip items unchanged since last poll

### Consider REST alternatives
- gh issue list uses GraphQL internally
- gh api /repos/{owner}/{repo}/issues uses REST (separate 5000/hour bucket)
- Could split load across both buckets

## Impact
Single-user with 3 repos was hitting GraphQL 0/5000 within an hour at 30s poll interval.

Closes jasonmadigan/clawdio#189


jasonmadigan:
[clawdio review]

CI is now passing. Here's my review:

---

## CI Status

build-and-test: **passing** (32s)

## Summary

Reworks GitHub polling and status checking to reduce API usage. Two main changes: (1) polling now fetches each repo once via REST `/repos/{owner}/{repo}/issues` instead of making separate GraphQL-backed `gh issue list` and `gh pr list` calls per query, with client-side filtering to apply query-specific constraints; (2) status sync supports batch checking via a `BatchStatusChecker` interface. Both changes move from GraphQL to REST, splitting load across separate rate limit buckets.

## Requirements (from #189)

- [x] Deduplicate polling -- merge queries hitting the same repos into a single fetch per repo
- [x] Fetch all open issues + PRs once, filter client-side
- [x] Batch status checks -- `BatchStatusChecker` interface + poller support
- [x] Consider REST alternatives -- all calls now use `gh api` (REST) instead of `gh issue/pr` subcommands (GraphQL)
- [x] Rate limit improvement -- moves from GraphQL (shared 5000/hr) to REST (separate 5000/hr bucket)

## Findings

### Important

1. **`BatchCheckStatus` doesn't actually batch** (`internal/adapter/github.go:146-172`). The method groups refs by repo but then calls `CheckStatus` individually for each. This is fine as a starting point -- it provides the interface and dedup-by-repo structure -- but the name oversells what it does. The issue mentioned "use a single GraphQL query to check multiple items" or "skip items unchanged since last poll". Neither is implemented yet. Worth noting in the PR body or a follow-up issue.

2. **`matchesQuery` label filter is case-sensitive** (`internal/adapter/github.go:387-397`) while assignee/author use `strings.EqualFold`. The old `gh issue list --label` was also case-sensitive (GitHub labels are case-sensitive), so this is correct behaviour, but the inconsistency with the assignee/author comparison stands out. Worth a comment or normalising the approach.

### Minor

3. **Extra blank line** at `internal/adapter/github.go:401` (between `matchesQuery` and `pollNotifications`).

4. **`resolveUsername` silently returns empty on error** (`internal/adapter/github.go:103-104`). If `gh api user` fails (auth issue, network), `@me` queries silently return no results. The old code delegated `@me` resolution to the `gh` CLI itself. Not a regression (the old behaviour would also fail), but a log line here would help debugging.

5. **Duplicate `BatchStatusChecker` interface** defined in both `internal/adapter/adapter.go:26-28` and `internal/platform/platform.go:25-27`. The poller uses the platform one; the adapter one appears unused. Consider removing the adapter copy to avoid confusion.

## Suggestions

- Add a follow-up issue for true batch status checking (e.g. fetching open issues for each tracked repo and cross-referencing, rather than N individual REST calls).
- Consider logging when `resolveUsername` fails so `@me` issues are visible.
- Remove the duplicate `BatchStatusChecker` in `internal/adapter/adapter.go` if nothing uses it.

## Verdict

The core goal is achieved -- polling is deduplicated, REST is used instead of GraphQL, and the structure is clean. The `BatchCheckStatus` naming is a bit misleading since it doesn't truly batch, but the interface is forward-looking and the poller integration is correct. Tests cover the new filtering logic well. Good work.

VERDICT: approve

PR #205, opened 2026-04-15
Bulk Ship It should show queued count when max_concurrent is reached

### LINKED ISSUE #194: Bulk Ship It should show queued count when max_concurrent is reached
## Problem

When selecting multiple items and clicking Ship, if the count exceeds `max_concurrent`, some items get queued silently. The UI gives no feedback that items are waiting -- they just stay in inbox while others move to in_progress. Looks like a bug to the user.

## Expected

After bulk Ship It, show feedback like "2 dispatched, 1 queued" via toast or inline in the bulk bar. The queued items should have a visual indicator (e.g. a "queued" badge) so it's clear they're waiting, not ignored.

## Current behaviour

- 3 items selected, Ship clicked
- 2 move to in_progress (max_concurrent = 2)
- 1 stays in inbox with no indication it's queued
- User thinks it's broken

jasonmadigan:
## Problem

When selecting multiple items and clicking Ship, if the count exceeds `max_concurrent`, some items get queued silently. The UI gives no feedback that items are waiting -- they just stay in inbox while others move to in_progress. Looks like a bug to the user.

## Expected

After bulk Ship It, show feedback like "2 dispatched, 1 queued" via toast or inline in the bulk bar. The queued items should have a visual indicator (e.g. a "queued" badge) so it's clear they're waiting, not ignored.

## Current behaviour

- 3 items selected, Ship clicked
- 2 move to in_progress (max_concurrent = 2)
- 1 stays in inbox with no indication it's queued
- User thinks it's broken

Closes jasonmadigan/clawdio#194


jasonmadigan:
[clawdio review]

Good. The `autopilot_dispatched` field tracks whether the item has actually been dispatched (workflow running), distinct from `autopilot_mode` which just means it's been flagged for autopilot. The badge logic correctly uses this: `autopilot_mode && !autopilot_dispatched` = queued.

Here's my review:

---

## CI Status

Passing -- `build-and-test` completed successfully.

## Summary

Adds user feedback when bulk Ship It encounters queuing due to `max_concurrent` limits. Shows a toast message with dispatched/queued counts and displays a "queued" badge on items that are waiting for a slot.

## Requirements (from #194)

- [x] Show feedback like "2 dispatched, 1 queued" via toast after bulk Ship It
- [x] Queued items should have a visual indicator ("queued" badge)
- [x] Queued items should be visually distinct from ignored/broken items

## Findings

**Important:**

1. **Toast for errors should use `showError`, not `showInfo`** (`App.tsx:340`). When items fail, `showInfo` gives the wrong visual signal. The error case should use `toast.showError(...)` to match severity.

2. **Mixed results (queued + errors) are not reported** (`App.tsx:337-341`). The `if/else if` means if there are both queued items *and* errors, only the queued toast shows. Consider combining: if both `queued > 0` and `errors > 0`, the message should reflect both.

**Minor:**

3. **No toast for all-success case** (`App.tsx:337-341`). If 3 items are selected and all 3 start successfully (no queuing, no errors), the user gets no toast at all. This is fine for single items, but for a bulk action some confirmation is arguably useful. Not a blocker -- the items visibly move to in_progress, so the feedback is implicit.

## Suggestions

For finding #2, something like:

```typescript
const parts: string[] = []
if (started > 0) parts.push(`${started} dispatched`)
if (queued > 0) parts.push(`${queued} queued`)
if (errors > 0) parts.push(`${errors} failed`)

if (parts.length > 0) {
  const msg = parts.join(', ')
  errors > 0 ? toast.showError(msg) : toast.showInfo(msg)
}
```

This handles all combinations in one pass and uses the right toast type.

## Verdict

The core feature works -- queued items get a badge and users see a toast. The two important findings are real but not critical; the badge (the more impactful part of this change) is correct. Request changes for the error toast type and the mixed-results gap.

VERDICT: changes_requested

jasonmadigan (commented):
Good. The three badge conditions are mutually exclusive — each covers a distinct combination of `autopilot_mode` and `autopilot_dispatched`. No overlap.

Now I have everything I need to write the review.

---

## CI Status

Passing — `build-and-test` completed successfully.

## Summary

Adds user feedback when bulk Ship It encounters queuing due to `max_concurrent` limits. Shows a toast message with dispatched/queued/failed counts and displays an amber "queued" badge on items waiting for a slot.

## Requirements (from #194)

- [x] Show feedback like "2 dispatched, 1 queued" via toast after bulk Ship It
- [x] Queued items should have a visual indicator ("queued" badge)
- [x] Queued items should be visually distinct from ignored/broken items

## Previous review

A previous review (posted as PR comment) raised two issues:
1. Toast for errors should use `showError`, not `showInfo`
2. Mixed results (queued + errors) silently dropped one category due to `if/else if`

Both addressed in `c6e9650`. The current code uses the `parts.join()` pattern suggested in the review and correctly picks `showError` vs `showInfo` based on whether any errors occurred.

## Findings

**Minor:**

1. **Single-item Ship It ignores return value** (`App.tsx:397`). The `onShipIt` handler calls `api.bulkShipIt([id]).then(() => loadItems())` — it discards the result array, so a single-item ship that gets queued shows no toast. This is pre-existing behaviour (not introduced by this PR), but now that the toast infrastructure exists, it'd be a one-line improvement. Not a blocker for this PR.

2. **Badge-queued and badge-skill-suggestion share identical colours** (`styles.css:2317-2324` vs `2325-2332`). Both use `--tint-amber-bg` / `--tint-amber-fg`. When a queued item also has a skill suggestion, you'd see two amber badges side by side. Unlikely in practice (queued items have already been dispatched so shouldn't show suggestions), but worth noting. Not a blocker.

## Verdict

The previous review's feedback has been addressed. The code is clean, the badge logic is correct (mutually exclusive conditions), the toast handles all result combinations, and backend status values (`started`, `queued`, `error`) match what the frontend filters for. Ship it.

VERDICT: approve

PR #213, opened 2026-04-16
Add pause autopilot, restart-from-step, fix skip for paused workflows

### LINKED ISSUE #164: Work item controls: add pause/resume autopilot, skip step, restart from step
## Current state

Several controls exist: reset to inbox, mark done, cancel autopilot, retry errored step, cancel workflow. Resume for paused workflows was recently added.

## Still missing

- **Skip step** -- advance the workflow past the current step without running it. Endpoint exists (`POST /api/work-items/{id}/skip-step`) but UI button is inconsistent.
- **Restart from specific step** -- restart a workflow from an arbitrary step, not just the current one. Useful when a later step needs re-running.
- **Pause autopilot** -- currently only cancel exists. Pause would keep the workflow run intact but stop auto-advancing. Resume would continue from where it paused.

## Context

Skills now own push/PR/review/merge. The workflow engine's `current_step` advances correctly after each session completes. Retry/resume dispatch the skill at `current_step`. Skip-step should increment `current_step` and dispatch the next skill.

jasonmadigan:
Closes #164

## Changes

**Pause autopilot** (`PUT /api/work-items/{id}/autopilot/pause`):
- Pauses the workflow run and disables auto-advance while keeping the run intact
- The current session continues but the workflow won't advance to the next step
- Resume via the existing start-autopilot endpoint which detects paused runs

**Restart from specific step** (`POST /api/work-items/{id}/restart-step`):
- Accepts `{ "step": N }` (0-based index) to restart the workflow from any step
- Works from errored, paused, or completed workflow runs
- Rewinds `current_step` and dispatches the skill at the target step
- Frontend: completed/errored steps show a "restart" hint on hover and are clickable

**Skip step fix**:
- Now works from both errored and paused workflow states (was errored-only)

**Frontend**:
- "Pause autopilot" button shown on running workflows with autopilot active
- Step list items are clickable when the workflow is stopped (errored/paused/completed)
- Hover reveals "restart" label on restartable steps


jasonmadigan (commented):
## CI

Passing.

## Summary

Adds three workflow control features: pause autopilot (keeps the run intact but stops auto-advance), restart from an arbitrary step, and skip-step support from paused state (was errored-only). Backend adds two new endpoints and broadens the skip-step handler. Frontend adds pause button on running workflows, clickable steps with hover "restart" hint, and supporting CSS.

## Requirements (from #164)

- [x] **Pause autopilot** -- `PUT /api/work-items/{id}/autopilot/pause` pauses the run and disables autopilot mode. Resume via existing start-autopilot endpoint which detects paused runs. UI shows "Pause autopilot" button on running workflows with autopilot active.
- [x] **Restart from specific step** -- `POST /api/work-items/{id}/restart-step` accepts `{ "step": N }`, validates bounds, rewinds `current_step`, dispatches. UI: completed/errored/paused steps are clickable with hover hint. Backend accepts errored, paused, or completed runs.
- [x] **Skip step from paused state** -- skip-step handler now accepts both `errored` and `paused` workflow runs. Error message updated accordingly.

## Findings

### Minor

1. **`handlePauseItemAutopilot` uses `ErrorDetail` for a non-error state** (`work_item_handlers.go:887`). Setting `error_detail = "paused by user"` on a paused run is semantically odd -- it's not an error. The frontend already shows "paused" status separately, and the error_detail block renders for any truthy value (line 829-831). This means a paused workflow shows "paused by user" in the error detail area. Not a bug, but slightly confusing UX -- pausing isn't an error. Consider clearing it or using a separate field if one exists. Low priority.

2. **`handleRestartFromStep` allows restarting from the current step on a completed run** -- the frontend excludes `i !== activeWorkflowRun.current_step` from `canRestart`, so the current step can't be clicked. But the backend has no such restriction. A direct API call with `step == current_step` on a completed run would re-run it. This is arguably fine (the retry-step endpoint already handles this case for errored/paused), just noting the frontend/backend parity gap.

3. **No confirmation on restart click** -- clicking a step immediately dispatches a restart. Since this rewinds workflow progress, an accidental click could re-run already-completed steps. Worth considering whether a confirmation would be prudent, though the current approach is consistent with the existing retry/skip buttons which also act immediately.

## Overall

Clean, well-scoped PR. The patterns follow the existing handlers closely (error rollback, profile loading, broadcast). Step bounds validation is correct. The frontend conditionals are sound -- `canRestart` correctly gates on non-running status and excludes the current step. CSS transitions are minimal and appropriate.

All three issue requirements are addressed. Would approve.

PR #215, opened 2026-04-16
Add light mode theme and theme switcher

### LINKED ISSUE #214: Light mode theme + theme switcher
## Overview

Add a light mode variant of the v3 design system and an easy way to switch between dark and light themes.

### Light mode palette
Design a light counterpart to the current dark theme. Key considerations:
- Backgrounds invert: white/grey surfaces instead of dark
- Text colours invert: dark text on light backgrounds
- Accent colours (indigo, emerald, amber, rose) stay the same but may need saturation/brightness adjustments for contrast on light backgrounds
- Borders become visible separators, not subtle overlays
- All existing contrast ratios must be maintained (WCAG AA minimum)

### Theme switcher
- Toggle in the sidebar (bottom, near settings) or in the topbar
- Three options: Light / Dark / System (follows OS preference via `prefers-color-scheme`)
- Preference persisted in localStorage
- Immediate switch, no page reload

### Implementation
- All colours already use CSS custom properties on `:root`
- Add a second set of token values under a `.theme-light` class or `[data-theme="light"]` attribute on `<html>`
- Scrollbar colours, shadows, and backdrop-filter values also need light variants
- Update `docs/style-guide.md` with the light palette

### Reference
- Current dark tokens: `docs/style-guide.md`
- Scrollbar styles need updating for light mode (currently tuned for dark backgrounds)

jasonmadigan:
Closes #214

## Changes

- **Light theme tokens:** Full `[data-theme="light"]` override block with inverted backgrounds, dark-on-light foreground, adjusted accents for WCAG AA contrast on white, and light-mode source badge colours.
- **Theme hook:** `useTheme` hook manages preference (dark/light/system), persists to localStorage, applies `data-theme` attribute on `<html>`, and tracks OS `prefers-color-scheme` changes when set to system.
- **Theme toggle:** Sidebar button (between spacer and activity) cycles dark -> light -> system. Icon changes to reflect current mode (moon/sun/monitor).
- **Tokenised interaction tints:** Extracted ~25 hardcoded `rgba(255 255 255 / ...)` hover/shadow values into `--hover-subtle/light/medium/strong`, `--shadow-card/elevated/modal/overlay`, `--scrim`, and `--grid-dot` tokens. Both themes override these so hover states, shadows, and overlays adapt correctly.
- **Scrollbar styles:** Light-mode scrollbar rules use dark track colours instead of white overlays.
- **Style guide:** Updated `docs/style-guide.md` with light palette tables, interaction tint reference, shadow tokens, and theme switcher docs.
- **Build fix:** Removed unused `ExternalLinkIcon` that was blocking `tsc -b`.


jasonmadigan (commented):
## CI Status

Build-and-test check is pending. No failures observed locally -- Go tests pass, and the changes are frontend-only.

## Summary

Adds a complete light theme via `[data-theme="light"]` on `<html>`, a `useTheme` hook for preference management (dark/light/system with localStorage persistence and OS media query listener), and a cycle button in the sidebar. Also introduces tokenised hover, shadow, scrim, and grid-dot custom properties in both themes, and migrates ~25 hardcoded `rgba(255 255 255 / ...)` values in component styles to the new tokens. Style guide updated.

Clean, well-structured work. The token extraction is the real value here -- it makes the codebase theme-aware by default rather than bolting light mode on as an afterthought.

## Requirements (from #214)

- [x] Light mode palette with inverted backgrounds, text, adjusted accents
- [x] Borders as visible separators (`rgba(0 0 0 / ...)` instead of white overlays)
- [x] Theme switcher in sidebar with Dark/Light/System
- [x] Preference persisted in localStorage
- [x] Immediate switch, no reload
- [x] Colours use CSS custom properties with `[data-theme="light"]`
- [x] Scrollbar colours updated for light mode
- [x] Shadow values updated for light mode
- [x] `docs/style-guide.md` updated with light palette

## Findings

### Minor

1. **Missed hardcoded value** (`styles.css:3610`): `.hm-step-dot` still uses `rgba(255 255 255 / 0.06)` instead of `var(--hover-strong)`. Will render as invisible white-on-white in light mode.

2. **FOUC on light theme** (`index.html`): When preference is `light` or `system` (resolving to light), the page renders dark-themed HTML until React mounts and `useTheme` calls `applyTheme`. A small inline `<script>` in `index.html` that reads localStorage and sets `data-theme` before the body renders would eliminate the flash. Not a blocker but noticeable on slower connections.

3. **Unrelated changes in this PR**: The `AgentFocus.css` grid column change (`1fr 320px 280px` to `1fr 30% 15%`) and the `ExternalLinkIcon` removal in `AgentFocus.tsx` aren't theme-related. The icon removal is fine (confirmed unused), but the grid change affects layout behaviour at different viewport widths. Might be better as a separate commit or at least called out in the PR description.

4. **Double blank line** (`AgentFocus.tsx:89-90`): Left behind by the `ExternalLinkIcon` removal.

### Observations (not blocking)

- The `accent-hover` in light mode is `#6366f1` which is actually the dark theme's base `--indigo`. Intentional (darker hover on lighter base) but worth double-checking contrast on the light surface.
- Scrollbar styles use hardcoded `rgba` values rather than the new `--hover-*` tokens. This is reasonable since scrollbar-color doesn't accept custom properties in all browsers, but noting for awareness.

## Verdict

The one genuinely problematic item is the missed `.hm-step-dot` hardcoded value which will be invisible in light mode. Everything else is minor polish. Given it's a single token swap, approving -- but please fix that step dot before merge.

jasonmadigan:
Addressed feedback: removed double blank line in AgentFocus.tsx (item 4). Items 1 and 2 (missed hardcoded rgba, FOUC) were already fixed in a90fc5b. Item 3 (unrelated grid/icon changes) noted but left as-is.

PR #234, opened 2026-04-17
fix: return pr_exists outcome for existing PRs

### LINKED ISSUE #165: Timeline: fix duplicate PR created events for existing PRs
## Backstory

Closed #135 addressed event ordering but the duplicate PR created event bug persists. The workflow engine returns a `pr_created:` outcome (engine.go lines ~353/373) even when the PR already exists, which adds another "PR created" event to the timeline.

## What's missing

When `push_and_pr` finds an existing open PR, it should return a different outcome (e.g. `pushed:` or `pr_updated:`) instead of `pr_created:`. The timeline then shows the correct event.

## What to do

- In the workflow engine, distinguish between "created new PR" and "found existing PR" outcomes
- Return `pr_updated` or `pushed` outcome when the PR already exists
- Update timeline rendering if needed to handle the new event type

jasonmadigan:
Closes #165

The `push_and_pr` action was returning `pr_created:url` for both new and existing PRs, causing duplicate "PR created" events in the timeline. Now returns `pr_exists:url` when an open PR is already found, so the timeline correctly distinguishes between creating a new PR and finding an existing one.

Changes:
- `engine.go`: two early-return paths now return `pr_exists:url` instead of `pr_created:url`
- `push_and_pr.yaml`: added `pr_exists` to outputs and transitions
- `manager.go`: handles `pr_exists:` prefix for PR URL extraction
- Timeline, AgentFocus, ActivityLog: added `pr_exists` event type with "PR updated" label
- Added `TestPostActionDonePRExistsAdvances` test


jasonmadigan (commented):
## CI Status

Passing (build-and-test).

## Summary

Introduces a `pr_exists` outcome in the workflow engine's `push_and_pr` action so that finding an existing open PR no longer emits a misleading `pr_created` event. The frontend timeline, activity log, and compact timeline all handle the new event type, displaying it as "PR updated".

## Requirements (from #165)

- [x] Distinguish between "created new PR" and "found existing PR" outcomes in the workflow engine
- [x] Return `pr_exists` outcome when the PR already exists (both the "item already tracks a PR" and "review already exists for branch" paths)
- [x] Update timeline rendering to handle the new event type

## Findings

### Minor

1. **Model comments not updated** (`internal/model/model.go:461,470`): The `EventType` doc comments on `WorkflowEvent` and `ActivityEntry` enumerate known event types but don't list `pr_exists`. These inline docs are how developers discover valid event types. Not blocking, but worth adding for consistency.

2. **No `recordEvent` call for `pr_exists` paths** (`engine.go:349,369`): When the engine returns `pr_exists`, it does not call `recordEvent` the way it does for `pr_created` (line 383). This is arguably correct behaviour since `pr_exists` means "nothing new happened, the PR was already there" and avoids polluting the timeline with redundant events. However, it means a `pr_exists` event will only appear in the timeline if something else records it. Currently the session manager handles the outcome to update the PR URL (manager.go:1470-1472), but it doesn't record a workflow event either. So `pr_exists` won't actually appear in the timeline at all as a discrete event. The frontend handles it if it arrives, but the backend never emits it as a workflow event. This might be the desired behaviour (silent PR URL update without timeline noise), but it's worth confirming the intent since the frontend work to render `pr_exists` events would then be dead code.

## Suggestions

- Add `pr_exists` to the event type comments in `model.go` for discoverability.
- Clarify the intended timeline behaviour: if `pr_exists` should appear as a timeline entry, add a `recordEvent` call; if it shouldn't, consider removing the frontend rendering for it (or keep it defensively for future use, and add a brief comment explaining why).

## Verdict

The core fix is sound: the engine no longer lies about creating a PR when one already exists. The skill YAML, workflow transitions, tests, and UI all handle the new outcome correctly. The test coverage is good. The minor points above are documentation/intent clarifications, not bugs.

Would approve, but GitHub prevents self-approval.

PR #235, opened 2026-04-17
feat: bidirectional agent-user messaging via MCP server

### LINKED ISSUE #198: Agent-user communication: bidirectional messaging and input requests
## Problem

Agents cannot communicate back to users. The UI has no working mechanism to send messages to running agents or receive input requests from them.

### What exists but doesn't work

- `POST /api/sessions/{id}/message` endpoint exists but the tmux pipe-pane setup is read-only (output only, no stdin pipe to the agent)
- `--input-format stream-json` is passed to Claude but bidirectional communication through tmux isn't wired
- The old Panel.tsx had a message input for `awaiting_input` state but it was non-functional
- Approve/deny endpoints exist for tool permissions but rely on the same broken stdin path

### What's needed

Agents need to be able to:
1. Ask the user a question and wait for a response
2. Request approval for an action and get yes/no
3. Surface important information that needs human attention

Users need to be able to:
1. Send a message to a running agent
2. Approve or deny tool use requests
3. Provide input when the agent asks for it

## Approach

Given the MCP direction (#197), the cleanest path is probably:

**Agent to user:** Agent calls a clawdio MCP tool like `request_user_input({ question: "...", options: [...] })`. The clawdio MCP server holds the request, broadcasts it to the UI via WebSocket. The UI shows the question with input/options. User responds, clawdio MCP server returns the response to the agent.

**User to agent:** For unsolicited messages, either:
- A clawdio MCP tool that the agent polls: `check_messages()` 
- Or keep the stdin pipe approach but fix it to actually work through tmux

**Tool approval:** Same MCP pattern. Agent's tool calls that need approval trigger `request_approval` on the clawdio MCP server, which holds until the user responds via UI.

## Depends on

- #197 (clawdio MCP server)

## Impact

Without this, agents run fully autonomously or not at all. There's no middle ground where a human can guide an agent during execution. This blocks the "draft_review" autonomy mode from working properly.

jasonmadigan:
Closes #198

## Summary

Adds an MCP server that enables bidirectional communication between agents and users. Agents can ask questions and wait for responses; users can send messages to running agents.

**Architecture:**
- `internal/mcp/` package: broker (request routing + response channels), server (JSON-RPC 2.0 over stdio), launcher (`--mcp-config` file generation)
- MCP server runs as a subprocess per agent session, communicates with the main server via HTTP long-poll
- Agent calls `request_user_input` tool, MCP subprocess POSTs to main server, blocks until user responds
- Agent calls `check_messages` to poll for unsolicited user messages

**Backend changes:**
- 5 new REST endpoints for the MCP subprocess and UI to communicate through
- Session manager wires `--mcp-config` into claude invocation
- WebSocket broadcasts `input_request` events for real-time UI updates
- Broker auto-cleans up on session cancel/completion

**Frontend changes:**
- Inline input request UI in AgentFocus (question text, option buttons, or freeform input)
- Message bar visible for all running sessions (sends via both stdin and MCP broker)
- New types and API methods for the messaging flow

## Test plan

- [x] All Go tests pass (`make test`)
- [ ] TypeScript type-check passes (`npx tsc --noEmit`)
- [ ] Manual test: dispatch an agent, verify MCP tools appear in claude's tool list
- [ ] Manual test: agent calls `request_user_input`, UI shows question, user responds, agent receives response
- [ ] Manual test: user sends message via input bar, agent picks up via `check_messages`


jasonmadigan (commented):
## CI status

Pending (build-and-test running).

## Summary

Implements bidirectional agent-user communication via MCP. The architecture is clean: a stdio-based MCP server subprocess per agent session, a broker for routing input requests and messages between the HTTP/WS layer and the MCP processes, and new REST endpoints + UI components for the interaction.

The approach matches what the issue described: `request_user_input` blocks via long-poll, `check_messages` drains a queue, WebSocket broadcasts input requests to the UI.

## Requirements (from #198)

- [x] Agent can ask the user a question and wait for a response (`request_user_input` tool)
- [x] Agent can surface information that needs human attention (via `check_messages` response)
- [x] User can send a message to a running agent (`send-message` endpoint + UI)
- [x] User can provide input when the agent asks (`pendingInputRequest` UI + `respond` endpoint)
- [ ] User can approve or deny tool use requests (not addressed -- issue listed this as a separate need, but the approach section suggested the same MCP pattern. Acceptable to defer if intentional)
- [x] MCP server subprocess wired into session lifecycle (config, cleanup)

## Findings

### Critical

1. **Closed channel returns empty string as success** (`broker.go:67-73`): When `CancelSession` closes the channel, `SubmitInputRequest` receives the zero value `""` from the closed channel and returns it as a successful response (nil error). The `select` should use the two-value receive `resp, ok := <-ch` and return an error when `ok` is false. This means cancelling a session while an input request is pending will send an empty string back to the agent as if the user responded.

### Important

2. **`AgentEvent.type` union missing `'input_request'`** (`ui/src/lib/types.ts:72`): The TypeScript `AgentEvent` type is `'agent_state' | 'tool_use' | 'text' | 'thinking' | 'control_request' | 'error'` but the code dispatches `"input_request"` events (`cmd/clawdio/main.go`). TypeScript won't catch this at compile time since `data` is `unknown`, but it means the type definition is inaccurate and consumers of the type will miss the new variant.

3. **No input validation on `sendAgentMessage`** (`session_handlers.go:236-253`): `handleSendAgentMessage` accepts any text including empty strings. An empty `body.Text` gets queued and delivered to the agent. Should reject empty/whitespace-only messages with a 400.

4. **Free-text input request reuses `messageText` state** (`AgentFocus.tsx:249-260`): The free-text input form for `pendingInputRequest` shares the `messageText` state with the general message input bar. Typing in one affects the other. If the user types a general message, then an input request appears, their in-progress text becomes the pre-filled response. Should use a separate state variable.

### Minor

5. **Unused `slog_` variable** (`mcp/server.go:18`): `var slog_ = logging.Component("mcp")` is declared but never used in `server.go`. It compiles because it's package-level, but it's dead code.

6. **`handleInputRequest` does not validate `body.ID`** (`session_handlers.go:148-166`): The handler accepts the `id` from the MCP subprocess body without checking it's non-empty. If the UUID generation in the MCP server fails silently or the JSON is malformed, an empty ID would be used as the key in the broker maps.

7. **`handleSendMessage` dual-path** (`AgentFocus.tsx:224-227`): The `Promise.allSettled` approach of sending via both the old stdin path and the new MCP broker path is pragmatic but should have a comment explaining why both paths are needed and when one can be removed.

## Suggestions

- For finding 1: change the select case to `case resp, ok := <-ch:` and return `ErrCancelled` (new sentinel error) when `!ok`.
- For finding 4: add a local `const [inputResponse, setInputResponse] = useState('')` for the input request form.
- Consider whether tool approval (#198 requirement 5) should be tracked as a follow-up issue or if it was intentionally deferred.

## Verdict

Requesting changes for the closed-channel bug (finding 1) -- it will cause incorrect behaviour when sessions are cancelled during pending input requests. The TypeScript type gap (finding 2) and shared state bug (finding 4) should also be fixed.

jasonmadigan:
Addressed feedback:

1. **Closed channel bug (critical):** `SubmitInputRequest` now uses two-value receive (`resp, ok := <-ch`) and returns `ErrCancelled` when the channel is closed by `CancelSession`.
2. **AgentEvent type:** Added `'input_request'` to the TypeScript union type.
3. **Empty message validation:** `handleSendAgentMessage` now rejects empty/whitespace-only messages with 400.
4. **Shared state bug:** Input request form now uses its own `inputResponseText` state, separate from `messageText`.
5. **Dead code:** Removed unused `slog_` variable from `mcp/server.go`.
6. **ID validation:** `handleInputRequest` now rejects empty `body.ID` with 400.
7. **Dual-path comment:** Clarified the stdin/MCP dual-path with removal guidance.

Tool approval (#198 requirement 5) was intentionally deferred -- it follows the same MCP pattern and can be added as a follow-up.

jasonmadigan (commented):
## CI status

## Summary

Follows up on the previous review. All seven findings have been addressed in `acfa8c6`:

1. **Closed channel bug (critical):** Fixed. `SubmitInputRequest` now uses two-value receive and returns `ErrCancelled` when the channel is closed. `ErrCancelled` sentinel added in `errors.go`.
2. **AgentEvent type union:** Fixed. `'input_request'` added to the TypeScript union on line 72 of `types.ts`.
3. **Empty message validation:** Fixed. `handleSendAgentMessage` rejects empty/whitespace-only text with 400.
4. **Shared state bug:** Fixed. `inputResponseText` is now a separate `useState` (line 100), independent of `messageText`.
5. **Unused `slog_`:** Removed from `server.go`.
6. **ID validation:** Fixed. `handleInputRequest` rejects empty `body.ID` with 400.
7. **Dual-path comment:** Added at line 225-226 of `AgentFocus.tsx` explaining the stdin/MCP dual path and when to remove the stdin path.

## Requirements (from #198)

- [x] Agent can ask user a question and wait for response (`request_user_input` MCP tool)
- [x] Agent can surface information needing attention (`check_messages` MCP tool)
- [x] User can send a message to a running agent (`send-message` endpoint + UI input bar)
- [x] User can provide input when agent asks (input request UI with options or free-text)
- [ ] Tool approval via MCP (intentionally deferred, same pattern applies later)
- [x] MCP subprocess wired into session lifecycle (config write, cleanup, broker integration)

## Minor observation

`TestBrokerCancelSession` verifies map cleanup but doesn't assert that `SubmitInputRequest` returns `ErrCancelled` -- the goroutine's return value is discarded. Not a blocker, but worth adding in a follow-up to exercise the new error path.

## Verdict

All previous findings resolved. Architecture is solid: broker mediates between MCP subprocesses and the HTTP/WS layer cleanly, the long-poll pattern for input requests is appropriate, and the UI integration handles both structured options and free-text responses well. Approve.

PR #236, opened 2026-04-17
feat: link workflow nodes to skill details with inline editing

### LINKED ISSUE #166: Workflow editor: link nodes to skill details and inline editing
## Backstory

Closed #108 asked for skill linking from workflow editor nodes and inline skill editing. The WorkflowBuilder component (676 lines) has full graph rendering and step inspection, but workflow nodes don't link to or open the corresponding skill. SkillsView has inline editing but it's not connected to the workflow editor.

## What's missing

- Clicking a workflow node's skill name should navigate to or open the skill detail view
- Inline skill editing from within the workflow editor (edit the skill template without leaving the workflow view)

## What to do

- Add `onSkillClick` handler to workflow nodes that opens the skill detail/editor
- Either navigate to SkillsView with the skill pre-selected, or embed a skill editor panel within the workflow builder

jasonmadigan:
Closes #166

## Summary

- Skill names in the workflow node inspector are now clickable links that navigate to the skills view with the skill pre-selected
- Added an inline skill editor panel within the workflow inspector showing contract details, post-actions, and an editable prompt template
- Added `/skills/:name` route support for deep-linking to specific skills
- Parallel branch skill names are also clickable links
- "Open full view" button navigates from the workflow editor to the full skill detail view

## Changes

- `useRouter.ts`: added `skillName` to router state, `/skills/:name` route parsing, `navigateToSkill` helper
- `App.tsx`: threads `navigateToSkill` to WorkflowBuilder and `skillName` to SkillsView
- `WorkflowBuilder.tsx`: clickable skill links in NodeInspector, inline skill editor panel with contract/prompt display and editing
- `SkillsView.tsx`: accepts `initialSkillName` prop, auto-selects the skill on mount
- `styles.css`: styling for skill links, action buttons, and inline editor panel


jasonmadigan (commented):
## CI status

Passing.

## Summary

Adds skill linking and inline editing to the workflow builder. Workflow node skill names become clickable links that navigate to the skills view with the skill pre-selected. An inline skill editor panel is embedded in the node inspector for editing skill content without leaving the workflow.

Changes across 5 files:
- **useRouter.ts**: adds `skillName` to router state, `navigateToSkill` helper, `/skills/:name` URL support
- **App.tsx**: threads `skillName` and `navigateToSkill` through to SkillsView and WorkflowBuilder
- **SkillsView.tsx**: accepts `initialSkillName` prop, auto-opens the skill on mount
- **WorkflowBuilder.tsx**: adds `onNavigateToSkill` prop, clickable skill links in inspector (both regular and parallel branches), inline skill panel with view/edit/save
- **styles.css**: styles for skill links, skill actions bar, and inline skill panel

## Requirements (from issue #166)

- [x] Clicking a workflow node's skill name navigates to or opens the skill detail view
- [x] Inline skill editing from within the workflow editor (edit skill template without leaving the workflow view)
- [x] Works for both regular steps and parallel branch skills

## Findings

No critical issues.

**Minor:**

1. **Stale data after inline save** (`WorkflowBuilder.tsx:658-665`): When the skill panel saves via `api.updateSkill`, the local `skills` state in WorkflowBuilder isn't refreshed. The prompt editor has the same pattern, so this is pre-existing rather than a regression, but worth noting -- a user who saves a skill inline and then opens the panel again will see stale content until they navigate away and back. Not blocking.

2. **`initialApplied` flag doesn't reset** (`SkillsView.tsx:90,107`): If the user navigates away from skills and back with a different `initialSkillName`, the flag stays `true` and the new skill won't auto-select. In practice this is unlikely since `navigateToSkill` sets a fresh `skillName` and the component would remount, but worth verifying if the component persists across view switches.

3. **HTML entity in JSX** (`WorkflowBuilder.tsx:593`): `&rarr;` is used directly in JSX. It works but `\u2192` or the literal `→` would be more idiomatic React. Cosmetic only.

## Verdict

Clean implementation. The router extension is well-structured -- `/skills/:name` URL support means these links are bookmarkable. The inline panel reuses existing CSS patterns and respects read-only state properly. The skill link behaviour covers both single-step and parallel-branch cases. Requirements from the issue are fully met.

Approving.

PR #245, opened 2026-04-18
feat: inject reflection prompt after repeated review/feedback cycles

### LINKED ISSUE #240: Reflection prompt after repeated review/feedback cycles
## Context

When review-pr requests changes and address-feedback fixes them, the cycle can repeat many times without progress. The agent keeps making the same fixes, the reviewer keeps finding the same issues. The workflow engine has a 50-transition safety valve but nothing that forces the agent to step back and think about why it's stuck.

From harness engineering research (Osmani): enforce `MAX_ITERATIONS` per agent with a forced reflection prompt asking "What failed? What specific change would fix it?"

## What to do

After N review/feedback cycles (suggest N=3) on the same workflow run, inject a meta-prompt into the next address-feedback session:

```
You have been through 3 review/feedback cycles on this PR.
Previous review findings that keep recurring: <list from session summaries>
Stop and think: what is the root cause? Are you fixing symptoms instead of the underlying problem?
If you cannot resolve this after one more attempt, output RESULT: nothing_to_do with an explanation of what's blocking you.
```

### Implementation

- Track cycle count in the workflow run (review-pr -> address-feedback = 1 cycle)
- In `session/manager.go` prompt building, check cycle count and inject reflection text
- The reflection prompt should include prior review summaries so the agent has context
- Consider also injecting the CI failure logs if available

### What this prevents

- Agents making cosmetic fixes that don't address the real issue
- Infinite loops where the same feedback keeps being raised
- Token waste on unproductive cycles

## References

- Osmani: "Loop guardrails with reflection" pattern
- Fowler: feedback sensors that "produce signals optimised for LLM consumption"

jasonmadigan:
Closes #240

After 3 review-pr sessions on the same workflow run, the next address-feedback dispatch gets a reflection section injected into its prompt. The section lists prior review findings and asks the agent to identify root causes instead of repeating surface-level fixes.

**Changes:**

- `prompt.go`: added `ReflectionContext` to `PromptOpts` and injection block in `BuildPrompt`
- `store.go`: added `ListWorkflowRunSessionsBySkill` to fetch prior review sessions with summaries
- `engine.go`: added `ReflectionThreshold` constant (3)
- `manager.go`: wired up cycle count check in `dispatchWorkflowStep` before dispatching address-feedback; added `Reflection` field to `DispatchOpts`

**Sequencing:** reflection fires at cycle 3, existing autopilot pause fires at cycle 4+ (`MaxReviewRounds+1`), giving the agent one informed attempt before we give up.

**Tests:** 5 new prompt tests (reflection present, absent, ordering, truncation) + 1 store test for the new query method.


jasonmadigan (commented):
## CI status

Passing.

## Summary

After 3 review-pr sessions on a workflow run, the next address-feedback dispatch gets a reflection section injected into its prompt. The reflection lists prior review findings and asks the agent to identify root causes before attempting another fix. Sequencing is correct: reflection fires at count 3 (`>=`), autopilot pause fires at count 4 (`>`), giving the agent one informed attempt.

## Requirements (from #240)

- [x] Track cycle count per workflow run (review-pr -> address-feedback)
- [x] Inject reflection prompt at threshold (N=3)
- [x] Include prior review summaries in the reflection
- [x] Escape hatch: `RESULT: nothing_to_do` instruction
- [x] Sequencing: reflection fires before autopilot pause
- [ ] CI failure log injection (issue listed this as "consider also" -- not blocking)

## Findings

No critical or important issues. Clean implementation that follows existing patterns.

**Minor:**

- `ReflectionThreshold` and `MaxReviewRounds` are both 3, and the relationship between them matters (reflection at `>=3`, pause at `>3`). The comment on `ReflectionThreshold` documents this, which is good. If either constant changes independently in future, the sequencing breaks silently. Acceptable for now since they're adjacent in `engine.go`.

- Errors from `CountWorkflowRunSessions` and `ListWorkflowRunSessionsBySkill` in the new reflection block are silently discarded. This is consistent with how the existing autopilot pause code handles `CountWorkflowRunSessions` at line 1347 -- the pattern is "fail open, skip reflection rather than block dispatch". Fine.

## Suggestions

None blocking. The CI failure log injection mentioned in the issue ("consider also") could be a follow-up.

## Verdict

Tests are thorough (5 prompt tests covering presence, absence, ordering, truncation, plus 1 store test). Code follows existing patterns. Sequencing between reflection and autopilot pause is correct and documented.

PR #256, opened 2026-04-21
fix: use wrapper script for tmux sessions to avoid ARG_MAX limits

### LINKED ISSUE #255: Session fails to start when prompt exceeds tmux command length limit
## Description

Sessions fail immediately with `tmux new-session: exit status 1: command too long` when the skill prompt is large. The prompt is passed as a CLI argument to `claude` inside `tmux new-session`, which hits OS `ARG_MAX` or tmux's own command length limits.

Triggered by #254 -- the issue body is ~10K characters, which expands to a 17,316 character prompt. Any issue with a large body will hit this.

## Steps to Reproduce

1. Create a GitHub issue with a long body (10K+ characters)
2. Dispatch a workflow against it (e.g. `ship-no-merge`)
3. Session errors immediately: `tmux new-session: exit status 1: command too long`

## Root Cause

`internal/tmux/tmux.go:NewSession()` passes the full `claude` command (including the prompt as a CLI argument) to `tmux new-session`. The prompt is constructed by expanding skill template variables (including `{{description}}` which contains the full issue body). When this exceeds the OS argument length limit, `exec.Command` fails.

```go
// internal/tmux/tmux.go:27-42
func NewSession(name, dir string, env []string, bin string, args []string) error {
    tmuxArgs := []string{"new-session", "-d", "-s", name, "-c", dir}
    for _, e := range env {
        tmuxArgs = append(tmuxArgs, "-e", e)
    }
    tmuxArgs = append(tmuxArgs, bin)
    tmuxArgs = append(tmuxArgs, args...)  // prompt is in here as a CLI arg
    cmd := exec.Command("tmux", tmuxArgs...)
    // ...
}
```

## Proposed Fix

Write the prompt to a temp file and pass it via a file-based mechanism instead of inlining as a CLI argument. Options:

1. **Write prompt to a temp file, pass path via `--prompt-file`** (if Claude CLI supports it -- needs checking)
2. **Write prompt to a temp file, use shell redirection** -- start tmux session with a shell command that reads the file: `tmux new-session -d -s name "claude --print < /tmp/prompt.txt"`
3. **Use `--prompt-file -`** with stdin piping via tmux `send-keys` after session start -- start session with just the `claude` binary args, then pipe the prompt via `send-keys` or `load-buffer`/`paste-buffer`

Option 1 is cleanest if the CLI supports it. Otherwise option 2.

The temp file should be cleaned up after the session starts (or on session completion).

## Acceptance Criteria

- [ ] Sessions start successfully with prompts of any reasonable size (tested with 20K+ character prompts)
- [ ] Prompt is not passed as a CLI argument to tmux/claude
- [ ] Temp files are cleaned up after use
- [ ] Existing short-prompt sessions continue to work (no regression)

## Review Guidance

- Core change in `internal/tmux/tmux.go` and `internal/session/manager.go`
- Check how the prompt is currently constructed and passed through to `NewSession`
- Verify cleanup path for temp files (both success and error cases)

## Manual Verification

1. Dispatch #254 (Pre-dispatch triage hook, ~17K prompt) with `ship-no-merge`
2. Verify session starts and agent begins working
3. Dispatch a normal short-description issue
4. Verify it also works (no regression)
5. Check `/tmp` for leftover prompt files after sessions complete

jasonmadigan:
Closes #255

## Problem

Sessions fail immediately with `tmux new-session: exit status 1: command too long` when the skill prompt is large (e.g. issues with 10K+ character bodies). The full `claude` command including the prompt is passed as arguments to `tmux new-session`, which hits OS `ARG_MAX` or tmux's own command length limits.

## Fix

Instead of passing the command and all its arguments directly to `tmux new-session`, write them to a temporary wrapper shell script and have tmux execute that script. The script self-deletes via `rm -f` before `exec`-ing the actual command, so no cleanup is needed from the caller.

The wrapper script approach:
- Avoids all argument length limits (the script is a file, not a CLI argument)
- Self-cleans on both success and error paths
- Uses POSIX `shellQuote` to safely handle any characters in the prompt (including single quotes, newlines, etc.)
- Works transparently for all callers -- `NewSession` signature is unchanged

## Changes

- `internal/tmux/tmux.go`: `NewSession` now writes a wrapper script via `writeWrapperScript` and passes `sh <script>` to tmux instead of the raw command
- `internal/tmux/tmux_test.go`: Added tests for `shellQuote`, `writeWrapperScript`, self-deletion, and a 30K argument stress test


jasonmadigan (commented):
## CI Status

**Failing.** `TestNewSessionLargeArgs` fails in CI:

```
--- FAIL: TestNewSessionLargeArgs (0.05s)
    tmux_test.go:113: session should exist with large args
```

The test creates a session running `echo` with a 30K argument. `echo` exits instantly, so by the time `SessionExists` checks, the tmux session has already terminated. This is a race condition. Use `sleep 30` (like `TestNewSessionAndKill` does) instead of `echo` -- the point is to verify the session *starts* with large args, not that `echo` handles them.

## Summary

Solid approach. Writing a self-deleting wrapper script avoids ARG_MAX limits cleanly. The `shellQuote` implementation is correct POSIX quoting, the self-deletion via `rm -f` before `exec` is safe (shell loads the full script before executing), and error paths clean up properly. No changes to the `NewSession` signature means zero caller impact.

## Requirements

From #255:

- [x] Sessions start successfully with prompts of any reasonable size (tested with 30K in `TestNewSessionLargeArgs`)
- [x] Prompt is not passed as a CLI argument to tmux/claude (wrapper script used instead)
- [x] Temp files are cleaned up after use (self-deleting script + `os.Remove` on error path)
- [x] Existing short-prompt sessions continue to work (`TestNewSessionAndKill`, `TestNewSessionWithEnv` unchanged)

## Findings

### Critical

1. **CI test failure** (`internal/tmux/tmux_test.go:96-115`): `TestNewSessionLargeArgs` uses `echo` as the session command. `echo` exits instantly, creating a race where `SessionExists` returns false because the session already terminated. Replace with `sleep 30` (consistent with other tests) and verify the session starts.

### Minor

2. **`PipePaneToFile` has the same quoting pattern this PR improves** (`tmux.go:172`): `PipePaneToFile` uses a raw `fmt.Sprintf("cat >> '%s'", filePath)` with a comment warning about single quotes. Not in scope for this PR, but worth noting as a follow-up -- could use the new `shellQuote` for consistency.

## Verdict

Fix the test to use `sleep 30` instead of `echo` so it doesn't race. Everything else is clean.

jasonmadigan:
Addressed feedback: fixed TestNewSessionLargeArgs race condition by using `sleep 30` instead of `echo` -- echo exits instantly so the session terminates before SessionExists checks it.

jasonmadigan (commented):
## CI Status

**Failing.** `TestNewSessionLargeArgs` still fails on the latest commit (`32a374a`):

```
--- FAIL: TestNewSessionLargeArgs (0.05s)
    tmux_test.go:114: session should exist with large args
```

The previous review correctly identified the `echo` race condition, but the fix (`sleep 30 <30K-string>`) doesn't work either. `sleep` rejects the second argument as an invalid time interval:

```
sleep: invalid time interval: xxxx...
```

The session exits immediately on both macOS and Linux because `sleep` doesn't accept arbitrary extra arguments. The session terminates before `SessionExists` can check it.

**Fix:** Use a command that ignores extra positional arguments. For example:

```go
err := NewSession(name, t.TempDir(), nil, "sh", []string{"-c", "sleep 30", "--", largeArg})
```

`sh -c 'sleep 30'` runs sleep in a subshell where positional args (`$1` etc.) are never referenced. The `--` and `largeArg` are passed as positional parameters to the shell but ignored. This tests that the wrapper script handles 30K of argument data without hitting ARG_MAX, while keeping the session alive.

## Summary

The core implementation is solid. The wrapper script approach cleanly avoids ARG_MAX limits, `shellQuote` is correct POSIX quoting, and the self-deletion via `rm -f` before `exec` is safe. The `NewSession` signature is unchanged so callers aren't affected. Error paths properly clean up the script file. Only the test is broken.

## Requirements

From #255:

- [x] Sessions start successfully with prompts of any reasonable size (tested with 30K in `TestNewSessionLargeArgs` -- test intent is correct, just needs a working command)
- [x] Prompt is not passed as a CLI argument to tmux/claude (wrapper script used instead)
- [x] Temp files are cleaned up after use (self-deleting script + `os.Remove` on error path)
- [x] Existing short-prompt sessions continue to work (`TestNewSessionAndKill`, `TestNewSessionWithEnv` unchanged and passing)

## Findings

### Critical

1. **CI test failure** (`internal/tmux/tmux_test.go:107`): `sleep 30 <30K-string>` fails because `sleep` rejects the non-numeric second argument. Use `sh -c 'sleep 30'` with the large arg as a positional parameter that's ignored.

### Minor

2. **Previously noted:** `PipePaneToFile` (`tmux.go:172`) uses raw single-quote wrapping that the new `shellQuote` could improve. Not in scope, agreed.

## Verdict

Fix the test command so CI passes. The implementation itself is clean.

jasonmadigan:
Addressed feedback: TestNewSessionLargeArgs now uses `sh -c 'sleep 30' -- <largeArg>` instead of `sleep 30 <largeArg>`. The shell ignores positional parameters while keeping the session alive, avoiding sleep's rejection of non-numeric arguments.

jasonmadigan (commented):
## CI Status

**Passing.** All checks green.

## Summary

Wraps the tmux session command in a self-deleting shell script to avoid OS ARG_MAX and tmux command length limits. Previously, the full `claude` command with all arguments (including potentially huge skill prompts) was passed directly as tmux arguments. Now `writeWrapperScript` writes them to a temp file, tmux executes `sh <script>`, and the script `rm -f`s itself before `exec`-ing the real command.

## Requirements

From #255:

- [x] Sessions start successfully with prompts of any reasonable size -- `TestNewSessionLargeArgs` passes a 30K argument
- [x] Prompt is not passed as a CLI argument to tmux/claude -- wrapper script used instead
- [x] Temp files are cleaned up after use -- script self-deletes via `rm -f` before `exec`; error path in `NewSession` calls `os.Remove`; error paths in `writeWrapperScript` clean up on write/close failures
- [x] Existing short-prompt sessions continue to work -- `TestNewSessionAndKill`, `TestNewSessionWithEnv` unchanged and passing

## Findings

No issues. The implementation is clean:

- `shellQuote` uses correct POSIX single-quote escaping (`'` -> `'"'"'`), tested with embedded quotes and empty strings
- Self-deletion before `exec` is safe -- `sh` loads the full script into memory before executing
- Error paths all clean up the temp file (three separate `os.Remove` calls covering write failure, close failure, and tmux failure)
- `NewSession` signature unchanged, single production caller (`session/manager.go:642`) unaffected
- Test suite covers: quoting correctness, script content, self-deletion behaviour, 30K arg stress test

## Previous review feedback

Both prior reviews flagged the test command -- first `echo` (exits instantly, race), then `sleep 30 <largeArg>` (sleep rejects non-numeric arg). Current version uses `sh -c 'sleep 30' -- <largeArg>` which correctly keeps the session alive while exercising the large argument path. Both issues resolved.

The `PipePaneToFile` quoting note (minor, not in scope) was acknowledged and left for follow-up. Fine.

**Verdict: approve** (posted as comment because GitHub blocks self-approval)

PR #275, opened 2026-04-21
feat: file-based agent heartbeats for crash detection

### LINKED ISSUE #246: Agent heartbeats (file-based crash detection)
## Summary

Add file-based heartbeat per worktree so the orchestrator can detect crashed or stuck agent sessions. The session manager writes a JSON heartbeat file periodically; if the heartbeat goes stale, the agent is assumed dead.

Based on gastown's heartbeat system (`internal/polecat/heartbeat.go`).

## Motivation

Currently, clawdio infers agent state from the Claude Code event stream. If the stream stops (process crash, hang, OOM), there's no detection mechanism -- the session stays in `running` state until the user notices.

File-based heartbeats decouple liveness detection from the event stream. They survive server restarts (files persist on disk) and work even when the event stream is broken.

## Proposed Solution

Write a JSON heartbeat file per session at `<worktree>/.clawdio-heartbeat.json`:

```json
{
  "timestamp": "2026-04-20T14:00:00Z",
  "state": "working",
  "context": "editing internal/session/manager.go",
  "session_id": "uuid"
}
```

**States:** `working`, `idle`, `stuck`, `exiting`

**Write frequency:** every 30-60 seconds from the session manager goroutine that reads the event stream.

**Staleness threshold:** configurable, default 3 minutes (matching gastown's `SessionHeartbeatStaleThreshold`).

**Design notes from gastown worth adopting:**
- Agent self-reports state; orchestrator only checks freshness
- Best-effort writes (errors silently ignored -- heartbeat is a signal, not a correctness requirement)
- Backwards-compatible: sessions without heartbeat files are not treated as stale (avoids false positives during rollout)

## Acceptance Criteria

- [ ] Session manager writes heartbeat file on session start and periodically during execution
- [ ] Heartbeat includes timestamp, state, context, and session ID
- [ ] Heartbeat file is cleaned up on normal session completion
- [ ] `ReadHeartbeat()` and `IsStale()` functions exist for the watchdog to consume
- [ ] Staleness threshold is configurable via config.json

## Review Guidance

- Core change is in `internal/session/` -- the heartbeat write loop
- Check that heartbeat writes don't block the event stream reader
- Verify cleanup on both normal completion and error paths

## Manual Verification

1. `make dev-api` and dispatch a work item
2. Check `~/.clawdio/worktrees/<item>/` for `.clawdio-heartbeat.json`
3. Verify timestamp updates periodically
4. Kill the claude process manually; verify heartbeat goes stale after threshold

## Context

From [gastown research](https://github.com/gastownhall/gastown) -- prerequisite for session watchdog.

jasonmadigan:
Closes #246

Adds file-based heartbeat per worktree so the orchestrator can detect crashed or stuck agent sessions.

**Changes:**

- `internal/session/heartbeat.go`: `Heartbeat` type, `WriteHeartbeat`, `ReadHeartbeat`, `IsStale`, `RemoveHeartbeat`. Atomic writes via temp file + rename. Best-effort (errors silently ignored).
- `internal/session/manager.go`: `startHeartbeatLoop` writes heartbeat every 30s with latest agent state detail. Wired into `runTmux`, `runBare`, and `resumeTmuxSession`. Cleanup on session completion. `cleanupStuckSessions` now checks heartbeat staleness for running sessions and kills hung agents.
- `internal/config/config.go`: `heartbeat_stale_threshold` field (duration string, default 3m).
- `internal/session/heartbeat_test.go`: Tests for write/read, staleness, cleanup, overwrites, error paths.

**Heartbeat file format** (`<worktree>/.clawdio-heartbeat.json`):
```json
{"timestamp":"2026-04-21T14:00:00Z","state":"working","context":"editing main.go","session_id":"uuid"}
```

Sessions without heartbeat files are not treated as stale (backwards-compatible).


jasonmadigan:
**triage:** This PR is already implemented and merged to its branch. It closes #246 which has clear acceptance criteria. The PR is mergeable with clean merge state, has one commit, no review comments, and the implementation matches the issue's requirements.

This is a completed implementation needing review. The work item is the PR itself, and it's ready for the review step of the ship workflow.

```
READINESS: ready
WORKFLOW: ship
```

- **readiness:** ready
- **workflow:** ship

jasonmadigan (commented):
## CI Status

## Summary

Adds file-based heartbeat per worktree so the orchestrator can detect crashed/stuck agent sessions independently of the event stream. Heartbeat file is written every 30s by a goroutine, and `cleanupStuckSessions` now checks staleness for sessions that are tracked in memory but may be hung.

## Requirements (from #246)

- [x] Session manager writes heartbeat file on session start and periodically during execution -- initial write in `startHeartbeatLoop`, then every 30s via ticker
- [x] Heartbeat includes timestamp, state, context, and session ID -- all four fields present in the `Heartbeat` struct
- [x] Heartbeat file is cleaned up on normal session completion -- `RemoveHeartbeat` called in defer blocks across `runTmux`, `runBare`, and `resumeTmuxSession`
- [x] `ReadHeartbeat()` and `IsStale()` functions exist for the watchdog to consume -- both exported in `heartbeat.go`
- [x] Staleness threshold is configurable via config.json -- `heartbeat_stale_threshold` field, parsed with `time.ParseDuration`

## Findings

**Minor:**

1. `docs/configuration.md` -- the new `heartbeat_stale_threshold` config field is not documented. Every other config field has a section there; this one should too. Worth adding a few lines explaining the field, its format (Go duration string), and the 3m default.

2. `heartbeat.go:32` -- the parameter `context` shadows the builtin `context` package name. Not a bug (the file doesn't import `context`), but mildly surprising if someone later adds an import. Low-priority but `ctx` or `detail` would be clearer.

3. `heartbeat.go:48` -- `os.Rename` return value is discarded. This is consistent with the best-effort design, but a failed rename after a successful temp-file write leaves a `.tmp` file behind. `RemoveHeartbeat` does clean up `.tmp` files, so this is fine in practice. Just noting it.

4. `manager.go:174` -- the staleness check only fires when `sess.WorktreePath != ""`. If a session somehow has an empty worktree path, staleness detection is silently skipped. This matches the "backwards-compatible" design from the issue ("sessions without heartbeat files are not treated as stale"), but worth confirming that's intentional for sessions with empty paths specifically (vs sessions where the heartbeat file simply doesn't exist).

## Suggestions

- Add a `heartbeat_stale_threshold` section to `docs/configuration.md` documenting the field, accepted format, and default. This is the only finding I'd call actionable; the rest are observations.

## Verdict

Clean implementation. Atomic writes, proper goroutine lifecycle (context-cancelled ticker, `<-done` for synchronous shutdown), test coverage of the core functions, and the integration into all three session modes is consistent. The `cleanupStuckSessions` enhancement is a nice bonus beyond the issue requirements.

The only gap is the missing docs entry, but that's not blocking -- it can be a follow-up.

PR #288, opened 2026-04-22
fix: link parallel branch sessions to convergence system

### LINKED ISSUE #287: Workflow engine can't dispatch parallel steps reached via goto
## Description

Two related bugs in the workflow engine's parallel step handling:

### Bug 1: Parallel steps reached via `goto` never dispatch

When address-feedback completes and the workflow follows a `goto` back to a parallel review step, the engine's `StepResult` returns `ParallelBranches` but the session manager never reads that field. The workflow gets stuck.

**Status:** Partially fixed -- `handleAdvanceResult` now dispatches `ParallelBranches`, but the dispatch doesn't link sessions to branch records (see bug 2).

### Bug 2: Parallel branch dispatch doesn't use convergence system

The fix for bug 1 dispatches parallel branches via `dispatchWorkflowStep`, which creates independent sessions not linked to `ParallelBranchRecord` entries. This means:
- Each branch session completes independently
- Each completion triggers `advanceWithContract` independently
- Each independently triggers address-feedback or merge
- Result: runaway loop -- 3 reviews -> 3 address-feedbacks -> 3 x 3 reviews -> exponential blowup

The correct flow should be:
1. `initParallelStep` creates `ParallelBranchRecord` entries in the DB (already works)
2. Each branch session must be linked to its branch record (session ID stored on the record)
3. When a branch session completes, it updates its branch record (already works for initial dispatch)
4. The convergence evaluator checks if all branches are done and the policy is satisfied (already works)
5. Only then does the engine advance to the next step

The gap is in step 2: `handleAdvanceResult` dispatches branches via `dispatchWorkflowStep` which doesn't set the session ID on the branch record. Without the link, the convergence evaluator never sees the sessions complete.

### Impact

Triggered a runaway loop that spawned 29+ concurrent sessions before being manually killed. The 50-transition cap didn't fire fast enough because each parallel fan-out counts as one transition but spawns 3 sessions.

## Root Cause

The session manager's `handleAdvanceResult` was added to dispatch `ParallelBranches` but naively -- it calls `dispatchWorkflowStep` per branch without linking sessions to `ParallelBranchRecord` entries. The convergence system requires this link to track branch completion.

## Fix Required

The parallel dispatch in `handleAdvanceResult` needs to:
1. Read the branch records created by `initParallelStep` (via `store.GetParallelBranches`)
2. Dispatch a session per branch, using the branch's workspace mode
3. Store the session ID on each branch record (via `store.UpdateParallelBranch`)
4. Ensure branch session completions route through `CompleteParallelBranch` (already works if session has correct workflow_run_id)

## Acceptance Criteria

- [ ] Parallel steps dispatch with sessions linked to branch records
- [ ] Convergence evaluator collects all branch verdicts before advancing
- [ ] `goto` to a parallel step works correctly (no runaway loops)
- [ ] `ship-reviewed` workflow completes end-to-end
- [ ] Parallel step round-trips (review -> feedback -> goto review) work

jasonmadigan:
Closes #287

Two bugs prevented parallel steps reached via `goto` from working correctly, causing runaway session spawning (29+ sessions observed).

**Bug 1:** `advanceWorkflow` never checked if a completing session belonged to a parallel branch. Each branch session triggered independent advancement instead of routing through `BranchCompleted` for convergence evaluation. This caused N branches to each independently trigger address-feedback or merge.

**Bug 2:** `handleAdvanceResult` dispatched parallel branches via `dispatchWorkflowStep`, which creates sessions not linked to `ParallelBranchRecord` entries. Without the link, `GetParallelBranchBySession` returns nil and convergence never evaluates.

**Changes:**

- `advanceWorkflow` checks `IsParallelSession` and routes through new `advanceParallelBranch` method which calls `BranchCompleted` for convergence
- `handleAdvanceResult` dispatches via new `dispatchParallelBranch` and calls `UpdateBranchSession` to link session IDs to branch records  
- `initParallelStep` cleans up stale branch records before creating new ones (needed for goto re-entry after address-feedback)
- `findStep` matches parallel steps by branch skill name (enables `goto: review-correctness` to find the parallel step)
- `DispatchWorkflow` handles the case where the first step is parallel
- New store method `DeleteParallelBranchesForStep` for targeted cleanup
- Tests for stale record cleanup and full goto round-trip (parallel -> convergence -> address-feedback -> goto parallel)


jasonmadigan:
**triage:** This is a pull request (#288), not an issue. The triage skill explicitly excludes PRs — they route directly to review-pr, address-feedback, or merge-pr.

The PR is already implemented with code changes. It needs review, not triage.

READINESS: ready
WORKFLOW: ship-reviewed

- **readiness:** ready
- **workflow:** ship-reviewed

jasonmadigan (commented):
## CI Status

build-and-test: **pass** (35s)

## Summary

Fixes two bugs in parallel step handling that caused runaway session spawning (29+ sessions). The first bug: `advanceWorkflow` didn't recognise parallel branch sessions, so each branch independently triggered advancement instead of routing through convergence. The second bug: `handleAdvanceResult` dispatched branches via `dispatchWorkflowStep` which doesn't link sessions to `ParallelBranchRecord` entries, so convergence never evaluated.

The fix adds `advanceParallelBranch` to route branch completions through `BranchCompleted`, creates `dispatchParallelBranch` that returns the session for linking, cleans up stale branch records on goto re-entry, and extends `findStep` to match parallel steps by branch skill name.

## Requirements (from #287)

- [x] Parallel steps dispatch with sessions linked to branch records (`handleAdvanceResult` now calls `UpdateBranchSession`)
- [x] Convergence evaluator collects all branch verdicts before advancing (`advanceParallelBranch` routes through `BranchCompleted`)
- [x] `goto` to a parallel step works correctly (`initParallelStep` cleans stale records, `findStep` matches branch skills)
- [x] Parallel step round-trips work (test: `TestParallelGotoRoundTrip`)
- [ ] `ship-reviewed` workflow completes end-to-end -- no integration test or evidence of end-to-end run

## Findings

### Critical

**`UpdateBranchSession` failure silently leaves branches unlinked** (`manager.go:1637-1639`)

If `UpdateBranchSession` fails, the error is logged but the loop continues. The session is already dispatched but not linked to its branch record. This is exactly the condition that caused the original runaway loop -- `BranchCompleted` calls `GetParallelBranchBySession`, gets nil, and returns an error at line 787-788 of `engine.go`. The branch session completes, `advanceWorkflow` calls `IsParallelSession` which also returns false (no linked record), so it falls through to the normal advancement path -- reintroducing the runaway behaviour for that branch.

This should be a hard failure that errors the workflow, same as the dispatch failure on the line above:

```go
if err := m.wfEngine.UpdateBranchSession(...); err != nil {
    wflog.Error("link branch session", ...)
    m.markWorkflowErrored(&run, fmt.Sprintf("link branch %d: %v", branch.BranchIndex, err))
    return
}
```

### Important

**No review-loop protection on the parallel path** (`manager.go:1558-1608`)

`advanceWithContract` (the non-parallel path) has `MaxReviewRounds` protection that pauses the workflow after too many changes_requested cycles (`manager.go:1506-1521`). `advanceParallelBranch` posts reviews and routes through `BranchCompleted` but has no equivalent guard. A parallel review step that keeps rejecting could loop indefinitely through `goto` -> parallel review -> address-feedback -> goto, since the per-transition cap (`MaxTransitions = 50`) counts transitions not sessions. With 3 parallel branches per round, you'd hit 50 transitions after ~16 rounds but spawn ~48 sessions in the process.

Suggestion: add the same `MaxReviewRounds` check in `advanceParallelBranch` before posting the review, or ensure the convergence path inherits the protection.

### Suggestion

**Synthetic session in `DispatchWorkflow` has empty fields** (`manager.go:1346`)

`synth := model.Session{WorkflowRunID: run.ID}` is passed to `handleAdvanceResult` which uses `sess.WorkflowRunID` -- this works. But if `handleAdvanceResult` ever grows to use other session fields (e.g. `WorkItemID` for error reporting), the synthetic session would silently produce zero-value lookups. Consider populating `WorkItemID` as well since `item` is in scope:

```go
synth := model.Session{WorkflowRunID: run.ID, WorkItemID: item.ID}
```

## Positive observations

- `TestParallelGotoRoundTrip` is a thorough integration test that exercises the full cycle: parallel review with rejection, address-feedback, push action, goto back to parallel, and verifies stale record cleanup. Good coverage of the exact scenario that caused the bug.
- The stale record cleanup in `initParallelStep` via `DeleteParallelBranchesForStep` is a clean approach -- deleting only records for the specific step rather than all records for the run avoids interference with other parallel steps if the workflow has multiple.
- `findStep` matching by branch skill name is the right way to resolve `goto: review-correctness` landing on a parallel step. The three-pass approach (skill -> name -> branch skill) maintains backwards compatibility.

## Verdict

The `UpdateBranchSession` silent failure is a gap that could reintroduce the exact runaway loop this PR aims to fix. That needs to be a hard error before merge. The review-loop protection gap is less urgent but worth addressing since the whole PR is about preventing runaway sessions.

jasonmadigan:
Addressed feedback:

- **Critical (UpdateBranchSession silent failure):** Now a hard error -- if linking fails, the workflow is marked errored and dispatch stops. Prevents unlinked sessions from falling through to the normal advancement path.
- **Important (MaxReviewRounds on parallel path):** Added the same changes_requested cycle guard to advanceParallelBranch, counting per branch skill name. Pauses autopilot after MaxReviewRounds cycles, matching the non-parallel path behaviour.
- **Suggestion (synthetic session fields):** Populated WorkItemID on the synthetic session in DispatchWorkflow.

jasonmadigan (commented):
## CI Status

build-and-test: **pass** (39s)

## Summary

Re-review after address-feedback round. The previous review raised three findings: one critical (UpdateBranchSession silent failure), one important (no MaxReviewRounds on parallel path), and one suggestion (synthetic session empty fields). All three are resolved.

## Previous findings -- verification

### Critical: UpdateBranchSession silent failure

**Fixed.** `manager.go:1654-1658` now calls `markWorkflowErrored` and returns on failure, matching the dispatch failure pattern above it. This closes the gap where an unlinked session could fall through to normal advancement and reintroduce the runaway loop.

### Important: MaxReviewRounds on parallel path

**Fixed.** `manager.go:1573-1587` adds the same `changes_requested` cycle guard to `advanceParallelBranch`. The parallel version counts by `sess.SkillName` rather than hardcoding `"review-pr"`, which is actually an improvement over the non-parallel path -- it correctly handles any review skill name.

### Suggestion: Synthetic session empty fields

**Fixed.** `manager.go:1346` now includes `WorkItemID: item.ID`.

## Requirements (from #287)

- [x] Parallel steps dispatch with sessions linked to branch records (`handleAdvanceResult` calls `UpdateBranchSession` with hard error on failure)
- [x] Convergence evaluator collects all branch verdicts before advancing (`advanceParallelBranch` routes through `BranchCompleted`)
- [x] `goto` to a parallel step works correctly (`initParallelStep` cleans stale records, `findStep` matches branch skills)
- [x] Parallel step round-trips work (test: `TestParallelGotoRoundTrip`, `TestInitParallelStepCleansStaleRecords`)
- [ ] `ship-reviewed` workflow completes end-to-end -- no integration test evidence, but the unit test coverage of the round-trip is thorough

## Five-axis assessment

**Correctness:** The three fixes are correct. The `UpdateBranchSession` error path now properly prevents the exact failure mode that caused the original runaway. The `MaxReviewRounds` guard uses per-skill counting which handles the multi-skill parallel case correctly.

**Readability:** `advanceParallelBranch` is well-structured -- it follows the same pattern as `advanceWithContract` with the parallel-specific routing added cleanly. The `markErrored` closure pattern is consistent with existing code.

**Architecture:** Parallel branch completion now has a proper dedicated path (`advanceParallelBranch`) rather than sharing the non-parallel `advanceWithContract`. This is the right call -- the convergence routing and per-branch review posting are fundamentally different from single-step advancement.

**Security:** No security concerns. Changes are internal workflow orchestration with no user input handling.

**Performance:** No concerns. The stale record cleanup (`DeleteParallelBranchesForStep`) is scoped to a specific step, avoiding unnecessary work.

## Positive observations

- The `TestParallelGotoRoundTrip` test is comprehensive: it exercises the full cycle of parallel review with rejection, address-feedback, push action, goto back to parallel, and verifies stale record cleanup. This is exactly the scenario described in the issue.
- The per-skill `MaxReviewRounds` counting in the parallel path is better than the hardcoded `"review-pr"` in the non-parallel path. Worth backporting that improvement eventually.

PR #292, opened 2026-04-22
feat: workflow step dependencies (needs) for DAG-based ordering

### LINKED ISSUE #248: Workflow step dependencies (needs)
## Summary

Add a `needs` field to workflow step definitions for DAG-based ordering instead of purely linear sequences. Steps whose dependencies are satisfied can run in any order or in parallel.

Based on gastown's formula step dependency model.

## Motivation

Current workflow steps execute in strict linear order (with `goto` for loops). This works for simple pipelines like `implement -> review -> merge`, but can't express more complex dependency graphs. For example:

- Security review and code review can run in parallel, but both must finish before merge
- Lint check doesn't depend on review, only on implement

The `needs` field lets you express these relationships declaratively.

## Proposed Solution

Add an optional `needs` field to workflow step YAML:

```yaml
steps:
  - id: implement
    skill: implement-issue
    then: push_and_pr

  - id: security-review
    skill: security-review
    needs: [implement]
    workspace: snapshot

  - id: code-review
    skill: review-pr
    needs: [implement]

  - id: merge
    skill: merge-pr
    needs: [security-review, code-review]
    on_approve: rebase_on_main
```

**Behaviour:**
- Steps with `needs` wait until all listed step IDs have completed
- Steps without `needs` and without unmet dependencies are eligible to run
- Multiple eligible steps run in parallel (reuses existing parallel step infrastructure)
- Steps still require an `id` field (currently optional for linear workflows)
- Backwards compatible: steps without `needs` execute in declaration order as today

**Engine changes:**
- Parse `needs` field during workflow loading
- Build dependency graph; validate no cycles
- `AdvanceWithContract` checks dependency satisfaction before dispatching next step
- Track completed step IDs on `WorkflowRun`

## Acceptance Criteria

- [ ] Workflow steps support optional `needs: [step_id, ...]` field
- [ ] Engine builds and validates DAG (cycle detection)
- [ ] Steps with satisfied dependencies run (potentially in parallel)
- [ ] Steps with unmet dependencies wait
- [ ] Backwards compatible with existing linear workflows
- [ ] At least one test workflow demonstrating parallel branches with convergence

## Review Guidance

- Core change in `internal/workflow/engine.go` (step eligibility logic)
- Check interaction with existing `goto` and convergence policies
- Verify cycle detection catches self-references and indirect cycles

## Manual Verification

1. Create a test workflow with parallel branches (e.g. two reviews that both `needs: [implement]`)
2. Dispatch a work item with this workflow
3. Verify both reviews start after implement completes
4. Verify the merge step waits for both reviews

## Context

From [gastown research](https://github.com/gastownhall/gastown) -- gastown's TOML formulas use `needs` for DAG-based step ordering.

jasonmadigan:
Closes #248

Steps can now declare `needs: [step_id, ...]` to express dependency graphs instead of purely linear sequences. Steps whose dependencies are all satisfied are dispatched, potentially in parallel.

## Changes

- **Model:** Add `id` and `needs` fields to `WorkflowStep`. Add `completed_steps` tracking to `WorkflowRun` (persisted as JSON in SQLite). Replaces the previous `name` field with the more general `id` field.
- **Parser:** Parse `id` and `needs: [a, b]` from workflow YAML.
- **Serialiser:** Round-trip `id` and `needs` fields.
- **Validation:** Duplicate ID detection, unknown needs references, self-references, and DAG cycle detection via Kahn's algorithm.
- **Engine:** `advanceToNext` marks completed step IDs and checks for DAG mode. `advanceDAG` finds all steps whose dependencies are satisfied and dispatches them (as parallel branches if multiple are ready). `findStep` and `resolveTransitionTarget` resolve by step ID first, then skill name.
- **Backwards compatible:** Workflows without `needs` execute linearly as before.

## Example

```yaml
steps:
  - id: implement
    skill: implement-issue
    then: push_and_pr

  - id: security-review
    skill: security-review
    needs: [implement]

  - id: code-review
    skill: review-pr
    needs: [implement]

  - id: merge
    skill: merge-pr
    needs: [security-review, code-review]
```

## Testing

- Parse tests for `id`/`needs` fields and round-trip
- Validation tests for cycles, unknown refs, self-refs, duplicates, missing IDs
- Engine tests for DAG advancement, dependency satisfaction, completed steps persistence
- All existing tests continue to pass (backwards compatibility)


jasonmadigan (commented):
## CI Status

Pending at time of review. Local `make test` passes (21 packages, all green).

## Summary

Adds `needs: [step_id, ...]` to workflow step definitions for DAG-based ordering. Steps without `needs` still run linearly. Steps with satisfied dependencies dispatch, potentially in parallel. Includes Kahn's algorithm cycle detection, persistence of completed step IDs in SQLite, and round-trip serialisation.

## Requirements (from #248)

- [x] Workflow steps support optional `needs: [step_id, ...]` field
- [x] Engine builds and validates DAG (cycle detection)
- [x] Steps with satisfied dependencies run (potentially in parallel)
- [x] Steps with unmet dependencies wait
- [x] Backwards compatible with existing linear workflows
- [ ] **At least one test workflow demonstrating parallel branches with convergence** -- no `needs`-based workflow was added to `workflows/`. The test fixtures in `engine_test.go` cover this in-memory, but the acceptance criterion asks for a real workflow file.

## Findings

### Critical

**DAG parallel branch completion doesn't mark individual step IDs as completed.**

When `advanceDAG` dispatches multiple ready steps (engine.go:266-292), it creates `ParallelBranchRecord` entries and returns `StepResult{ParallelBranches: ...}`. When these branches complete, they flow through `BranchCompleted` (engine.go:832), which calls `advanceToNext` -> `MarkStepCompleted(currentStep.ID)`. But `currentStep` is `wf.Steps[run.CurrentStep]`, and `run.CurrentStep` was never updated to reflect the individual branch steps -- it stays at whatever value it had before `advanceDAG` was called.

Concretely, with steps `[impl, security, review, merge]`:
1. `impl` completes -> `advanceDAG` dispatches `security` (index 1) and `review` (index 2) as parallel branches
2. `run.CurrentStep` is never set by `advanceDAG` for multi-branch dispatch (line 267 just sets phase)
3. When branches converge, `BranchCompleted` reads `wf.Steps[run.CurrentStep]` which is still `impl` (index 0)
4. The step it passes to `advanceToNext` is `impl`, not `security`/`review`
5. So `MarkStepCompleted` marks `impl` again (idempotent noop), but never marks `security` or `review`
6. `advanceDAG` then finds `security` and `review` ready again (infinite loop)

Additionally, `BranchCompleted` at line 853 checks `step.IsParallel()` and bails if false. DAG-dispatched branches are individual skill steps (not parallel steps), so this check would short-circuit to `Advance()` instead of evaluating convergence.

This needs careful design work -- the existing parallel infrastructure assumes a single parallel *step* with branches, not multiple independent steps grouped ad-hoc. The DAG multi-dispatch likely needs its own convergence path.

### Important

**1. `ReadySteps` skips root steps (steps without `needs`).** Lines 345-347 of model.go: if `len(step.Needs) == 0`, the step is skipped. This means `ReadySteps` never returns root steps. The `advanceDAG` method (engine.go:237-243) compensates by adding root steps separately, but this split logic is fragile -- any future caller of `ReadySteps` that doesn't replicate the root-step fix will get wrong results. `ReadySteps` should be self-contained: a step with no `needs` and not yet completed should be ready.

**2. `StepByID` is dead code.** Defined at model.go:323 but never called anywhere outside the declaration. `findStep` in engine.go is used instead. Remove it or use it.

**3. `decodeCompletedSteps` silently swallows corrupt JSON** (store.go:835). If the `completed_steps_json` column contains invalid JSON (migration bug, manual DB edit), the entire completed-steps list silently becomes nil. The workflow would re-run all steps from scratch. At minimum, log a warning.

### Suggestion

**1. Rename `name` -> `id` migration.** The PR renames `WorkflowStep.Name` to `WorkflowStep.ID` but doesn't add a migration or compat shim for workflows already stored with `name:` in YAML. Any workflow files using `name:` will silently lose their step identifiers after this change. The `ship-reviewed.yaml` update handles the one checked-in file, but user-authored workflows could break. Consider parsing `name` as a fallback alias for `id` in `parseStepKey`.

**2. Consistent workspace handling.** DAG branches are hard-coded to `WorkspaceShared` (engine.go:278). Existing parallel branch infrastructure allows per-branch workspace modes (`snapshot`, `separate`). If a DAG step's skill needs snapshot isolation (e.g. `security-review`), this isn't respected. The step's own workspace config (if any) should be forwarded.

## Positive Observations

- Kahn's algorithm for cycle detection is correctly implemented with clear separation between validation-time and runtime concerns.
- The test suite is thorough for the model and validation layers (cycle, self-ref, duplicate ID, unknown ref, needs-without-id, idempotent marking).
- The `Name` -> `ID` rename is a good simplification -- the distinction was always artificial.
- Backwards compatibility is well preserved: `UsesDAG()` gates the new code path cleanly.

## Verdict

The model layer, parser, serialiser, and validation are solid. The critical issue is in the engine's multi-step parallel dispatch path -- the convergence infrastructure doesn't support the DAG use case where multiple independent steps are grouped as ad-hoc parallel branches. This will cause incorrect behaviour (likely an infinite loop or convergence failure) when two or more DAG steps are dispatched simultaneously. The single-step DAG path (one ready step at a time) should work correctly.

Requesting changes for the critical convergence path issue before merge.

jasonmadigan:
**triage:** This is a pull request (#292), not an issue. The triage skill explicitly does not apply to PRs — they are routed directly to review-pr, address-feedback, or merge-pr.

This PR is already implemented and opened. It should be sent to a review workflow, not through triage.

READINESS: needs_input
QUESTIONS: This work item is a pull request, not an issue. PRs bypass triage and route directly to review skills (review-pr, security-review, address-feedback, merge-pr). The orchestrator should route PR #292 to the appropriate review workflow instead of triage.

- **readiness:** needs_input

**questions:**
- This work item is a pull request, not an issue. PRs bypass triage and route directly to review skills (review-pr, security-review, address-feedback, merge-pr). The orchestrator should route PR #292 to the appropriate review workflow instead of triage.

jasonmadigan:
Addressed feedback:

**Critical fix:** DAG parallel branch convergence now works. `BranchCompleted` detects DAG branches (individual steps dispatched as parallel branches) and routes them through `dagBranchCompleted`, which marks each step ID as completed and waits for all branches before calling `advanceDAG` for the next round. This prevents the infinite loop where `run.CurrentStep` never reflected the branch steps.

**Important fixes:**
- `ReadySteps` now includes root steps (no `needs`, not completed) -- callers no longer need to replicate root-step logic
- Removed `StepByID` dead code
- `decodeCompletedSteps` logs a warning on corrupt JSON

**Suggestions addressed:**
- Parser accepts `name:` as alias for `id:` for backwards compat with existing workflow files
- DAG branches forward the step's `workspace` config (new field on `WorkflowStep`) instead of hardcoding `WorkspaceShared`
- Added `ship-dag.yaml` workflow demonstrating parallel DAG branches with convergence

New tests: `TestDAGBranchConvergenceMarksSteps`, `TestDAGBranchWorkspaceForwarding`, `TestReadyStepsIncludesRootSteps`, `TestParseNameAsIDAlias`.

jasonmadigan (commented):
## CI Status

## Summary

Adds `needs: [step_id, ...]` to workflow step definitions for DAG-based ordering. Steps with satisfied dependencies dispatch (potentially in parallel), while steps with unmet dependencies wait. Includes Kahn's algorithm cycle detection, completed step persistence in SQLite, and serialiser/parser round-trip support. The previous review's critical convergence issue has been addressed with a dedicated `dagBranchCompleted` path.

## Requirements (from #248)

- [x] Workflow steps support optional `needs: [step_id, ...]` field -- `WorkflowStep.Needs` field, parser, serialiser all present
- [x] Engine builds and validates DAG (cycle detection) -- `detectNeedsCycles` via Kahn's algorithm in validate.go
- [x] Steps with satisfied dependencies run (potentially in parallel) -- `advanceDAG` dispatches multiple ready steps as parallel branches
- [x] Steps with unmet dependencies wait -- `ReadySteps` only returns steps whose needs are all in `completedSteps`
- [x] Backwards compatible with existing linear workflows -- `UsesDAG()` gates the new code path
- [x] At least one test workflow demonstrating parallel branches with convergence -- `workflows/ship-dag.yaml`

## Findings

### Important

**1. DAG errored branch behaviour.** In `dagBranchCompleted` (engine.go:930-961), when a branch errors, the pending count check at line 946 treats errored branches as "done" (the convergence proceeds). But the errored branch's step ID is still marked as completed via `MarkStepCompleted` at line 934, regardless of the branch's status. This means a step that failed is recorded as "completed", and downstream steps that depend on it will become eligible to run. For the current use case (security-review errors, merge still proceeds) this may be intentional, but it's worth documenting explicitly or adding a status check before marking the step completed.

**2. `ReadySteps` returns all root steps on every call, including already-dispatched ones.** When `advanceDAG` dispatches multiple parallel branches and one completes first, the re-entry into `advanceDAG` calls `ReadySteps` again. Root steps without IDs (steps where `step.ID == ""`) skip the `completed` check at line 333 and are returned as ready every time. In practice, all DAG steps should have IDs (validation enforces this for steps with `needs`), but root steps don't require IDs. A root step without an ID in a DAG workflow would be re-dispatched on every convergence round. Consider either requiring IDs on all steps when `UsesDAG()` is true, or tracking completed steps by index as a fallback.

### Suggestion

**1. `encodeCompletedSteps` ignores marshal error.** At store.go:826, `json.Marshal(steps)` result's error is discarded with `_`. For a `[]string`, this will never fail in practice, but it's a pattern worth avoiding -- a future refactor that changes the type could silently corrupt data. Handling the error (or adding a comment noting why it's safe to ignore) would be cleaner.

**2. `GetAllParallelBranches` could be scoped.** The new store method (store.go:908) fetches all branches for a workflow run regardless of step. For classic parallel steps that happen before or after DAG branches in a mixed workflow, this could return stale branch records from earlier parallel steps if `DeleteParallelBranches` wasn't called. Currently `dagBranchCompleted` calls `DeleteParallelBranches` on convergence, so this is safe for the intended flow, but worth noting as a future footgun.

**3. `goto` interaction with DAG.** In `advanceToNext` (engine.go:188-201), `goto` handling runs before the `UsesDAG()` check. A `goto` target in a DAG workflow will jump to a step by index and continue linearly, bypassing the DAG dependency graph entirely. This is probably fine (goto is for loops like the address-feedback cycle), but the semantics should be documented -- goto skips dependency checks intentionally.

## Positive Observations

- The `dagBranchCompleted` method is clean. Using `StepIndex` on the branch record to map back to the workflow step for marking completion is a solid approach that avoids the original convergence bug without adding new state.
- `ReadySteps` is now self-contained with root step inclusion, fixing the fragile split logic from the first iteration.
- The `name` -> `id` migration with `parseStepKey` accepting both keys is pragmatic backwards compatibility.
- Test coverage is thorough: convergence marking, workspace forwarding, root step inclusion, ID alias parsing, cycle detection, self-reference, duplicate IDs, needs-without-id, and idempotent marking are all covered.
- `ship-dag.yaml` is a clear, practical example that demonstrates the feature.

## Verdict

The previous review's critical issue (DAG convergence infinite loop) is properly fixed. The remaining findings are important but not blocking -- the errored-branch semantics and root-step-without-ID edge case are worth addressing but neither causes incorrect behaviour in the current supported workflows. Approving.

